PR #10, opened 2015-11-18
Adding e2e test.

### LINKED ISSUE #5: Add e2e tests
- Deploying with gcloud
  - local (--docker-build=local, and use `FROM php-nginx`, so that using local build)
  - remote (--docker-build=remote, upload the image with a different name)
- Web serving

tmatsuo:
fixes #5.

With this CL, we
- uploads a newly built image to gcr.io/php-mvm-a/php-nginx:testing if the functional tests passed
- run end to end tests against an app deployed with gcloud, and delete that version after the test
- if all the functional tests and end to end tests passed, and if it's a push on the master branch, it will upload the newly built image to gcr.io/php-mvm-a/php-nginx:latest

If all works well, we can deploy the new image automatically :)


tmatsuo:
Alright, the e2e test started working :)

@sjlangley Can you taka a look?

.travis.yml
```diff
@@ -24,6 +34,20 @@ before_install:
...
 - sudo mv php-cs-fixer /usr/local/bin
 - sudo chmod +x /usr/local/bin/php-cs-fixer
+# We want to fail fast for coding standard violations.
+- php-cs-fixer fix --dry-run --diff --level=psr2 --fixers=concat_with_spaces .
+- if [ ! -d ${HOME}/gcloud/google-cloud-sdk ]; then
+      mkdir -p ${HOME}/gcloud &&
+      wget https://dl.google.com/dl/cloudsdk/release/google-cloud-sdk.tar.gz --directory-prefix=${HOME}/gcloud &&
+      cd ${HOME}/gcloud &&
+      tar xzf google-cloud-sdk.tar.gz &&
+      printf '\ny\n\ny\ny\n' | ./google-cloud-sdk/install.sh &&
```
sjlangley: this seems pretty flaky - if it's not supported perhaps file a bug with cloud-sdk to add a --supress-prompts flag or something

dump_credentials.php
```diff
@@ -0,0 +1,27 @@
...
+ */
+
+/**
+ * Dumps the contents of the environment variable GOOGLE_CREDENTIALS_BASE64 to
+ * a file.
+ *
+ * To setup Travis to run on your fork, read TRAVIS.md.
+ */
+file_put_contents(
+    getenv('GOOGLE_APPLICATION_CREDENTIALS'),
```
sjlangley: not better to just print this and capture the output in the shell script? (don't know if travis lets you do that?)

.travis.yml
tmatsuo: Good point. Now I'll use some options and ask cloud-sdk team about this.

dump_credentials.php
tmatsuo: I think that is doable, but `gcloud auth activate-service-account` only accept file path.

.travis.yml
tmatsuo: Also there is an environment variable `CLOUDSDK_CORE_DISABLE_PROMPTS` and added it.

tmatsuo:
@sjlangley Thanks! I believe cloud sdk installation is now rock solid. PTAL

tmatsuo:
I forgot to parameterize  the project id in the end to end test and just fixed it. @sjlangley PTAL

sjlangley:
lgtm++

tmatsuo:
Thanks! Let's see if it can upload the image automatically :)

tmatsuo:
FYI, the image is successfully and automatically uploaded to gcr.io/php-mvm-a/php-nginx:latest which is great. Only thing is that the layers are entirely replaced because travis build environments are ephemeral. Ideally we can cache the layers as much as possible, so eventually I want to skip e2e tests in travis, and run e2e tests on somewhere else, where we can have persistent workspace.

PR #13, opened 2015-11-24
Test refactoring, added promote_by_default = false.

tmatsuo:


tmatsuo:
@sjlangley 

This PR is just a small refactoring, which makes jenkins configuration much simpler.

PR #16, opened 2015-12-10
[TASK] Modify build to use PHP 7.0.0 final

cedricziel:
This pull request modifies the build script to use the available PHP 7.0.0 final tar and verifies it.

It would be nice if you could merge and push it to the gcr registry.

I signed the CLA.

PR #17, opened 2015-12-10
[BUGFIX] Composer install should not disable functions

### LINKED ISSUE #15: Docker: Composer doesn't use supplied php.ini
The supplied composer runner doesnt respect the supplied php.ini which makes it really hard to run post-install scripts that utilize for example ``symfony/console`` to generate optimized autoloaders (example: Laravel 5.1)

```
[Symfony\Component\Process\Exception\RuntimeException]
The Process class relies on proc_open, which is not available on your PHP installation.
```

I see that you've disabled the function on purpose, but really they shouldnt be limited on the CLI for one-off commands.

From my point of view, there's multiple possible solutions to the problem:

1. use the supplied ``php.ini`` to allow save overrides
2. use another php.ini (for example ``php.cli.ini``) just for composer
3. use another ini setting on the cli
4. override the contained php.ini in the image

I know I can override the supplied ``php.ini`` - maybe just document it to make it easier for people who are not into docker very much?

cedricziel:
Adds a configuration directive to ``composer.sh`` to clear the disabled functions to
allow execution of composer post-install hooks that rely on native processes.

Closes #15

### DIFF
--- a/php-nginx/composer.sh
+++ b/php-nginx/composer.sh
@@ -58,6 +58,7 @@ EOF
         ${PHP_DIR}/bin/php \
         -d suhosin.executor.include.whitelist=phar \
         -d suhosin.executor.func.blacklist=none \
+        -d disable_functions= \
         /usr/local/bin/composer \
         install \
         --no-dev \


cedricziel:
The build fails as the RC8 tar is not available anymore.

This is fixed by #16

tmatsuo:
@cedricziel Alright. Let's merge #16 first. Then can you rebase?

cedricziel:
Sure-as soon as it's in :)

tmatsuo:
It's in now. Please rebase and push.

cedricziel:
Done. I will rebase #18 straight away.

tmatsuo:
Thanks! Merging :)

PR #20, opened 2015-12-11
[BUGFIX] PHP Binaries are not on $PATH

### LINKED ISSUE #19: PHP binaries are not on $PATH
In order to be able to execute arbitrary commands, the php binaries need to be on PATH. This applies to inheriting images as well, so it doesn't seem sufficient to export a modified PATH in the composer script, if an inheriting image would want to compile pecl extensions.

cedricziel:
This patch ensures that the currently used PHP binaries
are always on the PATH variable, so subsequent commands
inside the container can run arbitrary commands as well
and don't have to care about where this image puts the
binaries.

See: https://docs.docker.com/engine/articles/dockerfile_best-practices/#env
Closes #19

### DIFF
--- a/php-nginx/Dockerfile
+++ b/php-nginx/Dockerfile
@@ -52,7 +52,8 @@ ENV NGINX_DIR=/usr/local/nginx \
     OPENSSL_VERSION=1.0.1p \
     NGINX_VERSION=1.8.0 \
     PHP56_VERSION=5.6.16 \
-    PHP70_VERSION=7.0.0
+    PHP70_VERSION=7.0.0 \
+    PATH=/usr/local/php/bin:$PATH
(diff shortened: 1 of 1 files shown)


cedricziel:
The failing build could be a network flaw. I will build the image locally to see if there's anything I need to do.

cedricziel:
Okay. Locally it builds fine. Maybe the PGP Server had a hickup or stuff. :/

Can you please re-trigger the build on travis? It's only allowed for target repo owners.

JustinBeckwith:
Now that the build is passing, this looks good to me.  @tmatsuo?

tmatsuo:
LG, merging. Thanks for the contribution :)

PR #23, opened 2015-12-11
[DOC] Mention how to enable disabled functions

cedricziel:
Besides suhosin, the main php.ini also uses disable_functions
to disable functions.

Other than on the command line, the value of that directive
needs some string to override the default.

Also mentions that you effectively replace the php.ini
completely.

### DIFF
--- a/php-nginx/README.md
+++ b/php-nginx/README.md
@@ -60,13 +60,41 @@ is as follows:
 suhosin.executor.func.blacklist="escapeshellarg, escapeshellcmd, exec, highlight_file, lchgrp, lchown, link, symlink, passthru, pclose, popen, proc_close, prog_get_status, proc_nice, proc_open, proc_terminate, shell_exec, show_source, system, gc_collect_cycles, gc_enable, gc_disable, gc_enabled, getmypid, getmyuid, getmygid, getrusage, getmyinode, get_current_user, libxml_disable_entity_loader, parse_str, phpinfo, phpversion, php_uname"
 ```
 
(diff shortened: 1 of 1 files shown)


php-nginx/README.md
```diff
@@ -60,14 +60,42 @@ is as follows:
 suhosin.executor.func.blacklist="escapeshellarg, escapeshellcmd, exec, highlight_file, lchgrp, lchown, link, symlink, passthru, pclose, popen, proc_close, prog_get_status, proc_nice, proc_open, proc_terminate, shell_exec, show_source, system, gc_collect_cycles, gc_enable, gc_disable, gc_enabled, getmypid, getmyuid, getmygid, getrusage, getmyinode, get_current_user, libxml_disable_entity_loader, parse_str, phpinfo, phpversion, php_uname"
 ```
 
-To enable them, you can add `php.ini` file in the project directory
-and override the default settings above. Here is an example for only
-enabling `phpinfo()`.
+To enable them, you can add a `php.ini` file in the project root directory,
+which will override the default ini file completely. Here is an example for
+enabling only `phpinfo()`.
```
tmatsuo: `php.ini` in the project root directory is added with --with-config-file-scan-dir=$APP_DIR configuration option.

https://github.com/GoogleCloudPlatform/php-docker/blob/master/php-nginx/build_deps.sh#L163

It should just added to the default php.ini. Did you observe otherwise?

cedricziel: You're right - my fault :)

cedricziel:
Alright-that should be it.

PR #24, opened 2015-12-11
[FEATURE] Allow usage of a personal GitHub oAuth token

### LINKED ISSUE #22: [Q] General direction of the image
I have an important question while I try to bend this image to my needs: What is the feature-complete state of this image?

The heroku buildpack for php is a really great source of inspiration in terms of what is possible and needed for running modern, sophisticated php applications on the platform side of things. I am aware that it may slightly conflict with the "many tiny applications"-paradigm Google drives, but I'm curious if you are willing to go towards that direction with this image as well.

I'd like you to take a look at the compile script of the heroku buildpack for php to see some of the features it detects. (https://github.com/heroku/heroku-buildpack-php/blob/master/bin/compile)

I am aware that I can run a docker image with a heroku buildpack on managed vms - but apart from the general, language specific features, I would also expect AppEngine features from the sandbox to get integrated in the future with this image (CloudSQL with a simple socket). If it wasn't for those, I'd just use the one of the heroku flavoured docker images with managed vms.

cedricziel:
A new environment variable `COMPOSER_GITHUB_OAUTH_TOKEN` can
now be used to let composer authenticate against the GitHub API on
behalf of the user. This allows the builder VM to always use the
`dist` packages and never fall back to the source method.

See https://getcomposer.org/doc/articles/troubleshooting.md#api-rate-limit-and-oauth-tokens

### DIFF
--- a/php-nginx/composer.sh
+++ b/php-nginx/composer.sh
@@ -49,6 +49,23 @@ EOF
         ln -sf ${PHP56_DIR} ${PHP_DIR}
     fi
 
+    # Handle custom oauth keys (Adapted from https://github.com/heroku/heroku-buildpack-php/blob/master/bin/compile)
+    COMPOSER_GITHUB_OAUTH_TOKEN=${COMPOSER_GITHUB_OAUTH_TOKEN:-}
+    if [[ -n "$COMPOSER_GITHUB_OAUTH_TOKEN" ]]; then
+        if curl --fail --silent -H "Authorization: token $COMPOSER_GITHUB_OAUTH_TOKEN" https://api.github.com/rate_limit > /dev/null; then
+            ${PHP_DIR}/bin/php \
+              -d suhosin.executor.include.whitelist=phar \
+              -d suhosin.executor.func.blacklist=none \
+              -d disable_functions= \
+              /usr/local/bin/composer config -g github-oauth.github.com "$COMPOSER_GITHUB_OAUTH_TOKEN" &> /dev/null # redirect outdated version warnings (Composer sends those to STDOUT instead of STDERR)
+            echo 'Using $COMPOSER_GITHUB_OAUTH_TOKEN for GitHub OAuth.'
+        else
+            echo 'Invalid $COMPOSER_GITHUB_OAUTH_TOKEN for GitHub OAuth!'
+        fi
+    fi
+    # no need for the token to stay around in the env
+    unset COMPOSER_GITHUB_OAUTH_TOKEN
+
     # Remove the vendor directory if it exists.
     if [ -d ${APP_DIR}/vendor ]; then
         rm -rf ${APP_DIR}/vendor
--- a/php-nginx/README.md
+++ b/php-nginx/README.md
@@ -223,6 +223,19 @@ These extensions are only available with PHP 5.6:
 You can just have php.ini file in your project directory. This file is
 added to the main configuration file.
 
+## GitHub oAuth token
+
+If you see messages like `Failed to download vendor/package from dist: Could not
+authenticate against github.com`, it means your IP (or effectively the IP of
+the ComputeEngine builder VM when you deploy to AppEngine) exceeded the
+API rate limit for anonymous access and composer can now only use the
+(significantly slower) VCS source method for fetching packages. To fix it,
+you can add a environment variable `COMPOSER_GITHUB_OAUTH_TOKEN` to your
+deployment (`app.yaml` or `Dockerfile` are both fine).
+
+As this is your personal credential for GitHub, please make sure you do not
+commit this token to your repository.
+
 ## Deploy with gcloud
 
 You can deploy the app by:


php-nginx/composer.sh
```diff
@@ -49,6 +49,28 @@ EOF
...
+    if [[ -n "$COMPOSER_GITHUB_OAUTH_TOKEN" ]]; then
+        if curl --fail --silent -H "Authorization: token $COMPOSER_GITHUB_OAUTH_TOKEN" https://api.github.com/rate_limit > /dev/null; then
+            /usr/local/bin/composer config -g github-oauth.github.com "$COMPOSER_GITHUB_OAUTH_TOKEN" &> /dev/null # redirect outdated version warnings (Composer sends those to STDOUT instead of STDERR)
+            echo 'Using $COMPOSER_GITHUB_OAUTH_TOKEN for GitHub OAuth.'
+        else
+            echo 'Invalid $COMPOSER_GITHUB_OAUTH_TOKEN for GitHub OAuth!'
+        fi
+    else
+        # Don't forget to remove any stored key if it's gone from the env
+        /usr/local/bin/composer config -g --unset github-oauth.github.com &> /dev/null # redirect outdated version warnings (Composer sends those to STDOUT instead of STDERR)
```
tmatsuo: Is it failing because of suhosin?

cedricziel: Argh. Yes. Most probably. Maybe the arguments for composer should be stored in a variable to be included everywhere. I'll fix it.

php-nginx/composer.sh
```diff
@@ -49,6 +49,36 @@ EOF
...
+        else
+            echo 'Invalid $COMPOSER_GITHUB_OAUTH_TOKEN for GitHub OAuth!'
+        fi
+    else
+        # Don't forget to remove any stored key if it's gone from the env
+        ${PHP_DIR}/bin/php \
+          -d suhosin.executor.include.whitelist=phar \
+          -d suhosin.executor.func.blacklist=none \
+          -d disable_functions= \
+          /usr/local/bin/composer -g --unset github-oauth.github.com &> /dev/null # redirect outdated version warnings (Composer sends those to STDOUT instead of STDERR)
```
tmatsuo: composer config?

tmatsuo:
Re: `dist` and `src` methods

I still don't understand the token is needed in order to use the `dist` method. With the `dist` method, what does composer download and from where?

tmatsuo:
> I still don't understand the token is needed in order to use the dist method. With the dist method, what does composer download and from where?

`dist` will download from the url specified by the package's composer.json.

tmatsuo:
Oh, I closed it by mistake.

cedricziel:
Oookay - just chopped through the composer code and found the relevant piece of information - the local client connects to the github api to be able to [retrieve the checksum for the dist-zipfile](https://github.com/composer/composer/blob/master/src/Composer/Repository/Vcs/GitHubDriver.php#L134).

As that's an API call for _every_ single dependency in the lock-file (nested dependencies as well of course), this falls under the API rate-limiting of GitHub and can quickly get over anonymous API limit for the current IP.

Requesting it with curl (without a token present):

```
$ curl -v https://api.github.com/repos/composer/composer/zipball/master.zip

# ...headers before...
< X-RateLimit-Limit: 60
< X-RateLimit-Remaining: 59
< X-RateLimit-Reset: 1449862489
< Location: https://codeload.github.com/composer/composer/legacy.zip/master.zip
# ...headers after...
```

If the token is present, the composer client will add the appropriate Authorization header.

If the rate limit is exceeded, composer will do source downloads (=clone entire repositories to checkout a single commit) which prolongues the build process dramatically.

php-nginx/composer.sh
```diff
@@ -49,6 +49,36 @@ EOF
...
+        else
+            echo 'Invalid $COMPOSER_GITHUB_OAUTH_TOKEN for GitHub OAuth!'
+        fi
+    else
+        # Don't forget to remove any stored key if it's gone from the env
+        ${PHP_DIR}/bin/php \
+          -d suhosin.executor.include.whitelist=phar \
+          -d suhosin.executor.func.blacklist=none \
+          -d disable_functions= \
+          /usr/local/bin/composer config -g --unset github-oauth.github.com &> /dev/null # redirect outdated version warnings (Composer sends those to STDOUT instead of STDERR)
```
bshaffer: why are removing stored keys? Won't this be a new container, and so nothing will be set here?

php-nginx/composer.sh
```diff
@@ -49,6 +49,36 @@ EOF
...
+        if curl --silent https://api.github.com/rate_limit | python -c 'import sys, json; sys.exit((json.load(sys.stdin)["resources"]["core"]["remaining"] > 0))'; then # yes, check > 0, not < 1 - exit status of 0 will trigger the if
+            echo "You've reached the GitHub API's request rate limit.
```
bshaffer: Won't this error happen in composer by default, so a check here is redundant?

cedricziel: The check here is purely for diagnostic purposes, so the person performing the build can investigate and get a hint on slow builds. - I'll remove it if you think this is a bad idea.

php-nginx/composer.sh
cedricziel: I agree-until some kind of layer cache is used in an inheriting image. I would just leave this here to avoid side effects. What's your oppinion?

php-nginx/composer.sh
tmatsuo: The embedded python concerns me a bit too.

I think you get messages like:

```
Downloading: Connecting...    Failed to download jakub-onderka/php-console-color from dist: The "https://api.github.com/repos/user/repo/hash" file could not be downloaded (HTTP/1.1 403 Forbidden)
```

Isn't the message clear enough for developers to diagnose it?

cedricziel:
I removed the parts dealing with the non-existent token. Composer will display a message anyways.

tmatsuo:
LGTM
I would wait for @bshaffer

tmatsuo:
Ah I have one comment. Please wait

tmatsuo:
nm LGTM

tmatsuo:
Alright, I'm going to merge this. Thanks for the contribution!

PR #28, opened 2015-12-14
[FEATURE] Add experimental PHP7 memcached extension

cedricziel:


php-nginx/build_deps.sh
```diff
@@ -280,6 +281,18 @@ function build_php7 {
   strip ${PHP7_DIR}/bin/php ${PHP7_DIR}/sbin/php-fpm
 }
 
+function build_php7_memcached {
```
tmatsuo: Is it possible to do the same thing as PHP 5.6? We build the extensions with PHP itself.
Also can you add TODO to fetch the source code from pecl (stable version)?

cedricziel:
I added the TODO and switched to building it with the interpreter.

tmatsuo:
LGTM, merging

Issue #27, opened 2015-12-14
Applications cannot write to their own filesystem


cedricziel:
Booting a simple Laravel5 application results in errors, because it wants to cache the current container to ``$APP_ROOT/bootstrap/cache/services.json`` and ``file_put_contents`` fails.

Here's an application illustrating the issue: https://github.com/cedricziel/gae-docker-writable-issue

Just deploy it to AppEngine and open up the index route.

The underlying issue seems to be that though permissions seem correct, the application cannot write to the filesystem. While this could be fixed on build-time for this particular application, one of the promises managed vms make is a writable file system so applications caching files to disk don't have to jump through hoops.

cedricziel:
Revisiting the issue it's caused by [a directive in the Dockerfile](https://github.com/GoogleCloudPlatform/php-docker/blob/ad70cb50a289db6cf89d4e4314ee62c647d15563/php-nginx/Dockerfile#L99). I was surprised that the container overrides the filesystem modes originally set.

Do you intend to keep the file modes that restrictive?

sjlangley:
Writing to the filesystem is probably the most common path of exploits for PHP applications, so we configure the app folder as read-only be default.

I suggest configuring the cache folder somewhere outside of the application root.

cedricziel:
I'm well aware of the security implications - yet many php applications need one or more compile steps which can be invoked either lazily (on request time) or on build-time.

To save others from investigating about why on earth their file modes where overriden, I added a note to the images' README in #29.

Thank you for your answer :)

PR #29, opened 2015-12-15
[DOC] Mention read-only filesystem

cedricziel:
See #27

### DIFF
--- a/php-nginx/README.md
+++ b/php-nginx/README.md
@@ -281,6 +281,13 @@ Note: this doesn't reflect any additional web server configuration.
   provide an easier way to install extensions.
 - PHP 7 installation is still missing various extensions including
   suhosin.
+- As a security measure, the `/app` directory is changed to being read-only
+  during the build. If you need to write to specific folders in your `/app` 
+  directory, you need to make it writeable to the `www-data` user as a separate
+  run-step in your Dockerfile.
+  ```
+  RUN chmod -R 750 ${APP_DIR}/bootstrap/cache
+  ```
 
 ## What does this image do
 


php-nginx/README.md
```diff
@@ -281,6 +281,13 @@ Note: this doesn't reflect any additional web server configuration.
   provide an easier way to install extensions.
 - PHP 7 installation is still missing various extensions including
   suhosin.
+- As a security measure, the `/app` directory is changed to being read-only
+  during the build. If you need to write to specific folders in your `/app` 
+  directory, you need to make it writeable to the `www-data` user as a separate
+  run-step in your Dockerfile.
```
tmatsuo: I actually prefer using directory outside of `/app` or at least outside of the webroot, rather than changing the default permissions.

What do you think?

cedricziel: Hm. I totally see your arguments. - I would prefer it as well. But I also see what others do :)

IMHO the use case is more like the one I mentioned in #27: People use a framework, switch the document root anyways and just need the app to run the way it is intended to run. To accomplish this, the framework needs to be able to write into 1 or 2 locations. If it cannot, that's a bummer.

Besides that: a writable directory is a writable directory - wether it's in `/app` or not. - The key is the document root switch. Don't you think so?

tmatsuo: If you still have a writable directory inside your document root, it's still a big problem.

cedricziel: Of course it should be outside the document root, but inside the application root. Sorry if I wasn't explicit enough about it.

I will amend the commit to be more explicit about it.

tmatsuo:
@cedricziel Do you still want to amend?

cedricziel:
Alright, thank you :)

PR #33, opened 2015-12-17
Upgraded to PHP 7.0.1.

### LINKED ISSUE #32: Upgrade to PHP 7.0.1

tmatsuo:
Fixes #32

### DIFF
--- a/php-nginx/Dockerfile
+++ b/php-nginx/Dockerfile
@@ -52,7 +52,7 @@ ENV NGINX_DIR=/usr/local/nginx \
     OPENSSL_VERSION=1.0.1p \
     NGINX_VERSION=1.8.0 \
     PHP56_VERSION=5.6.16 \
-    PHP70_VERSION=7.0.0 \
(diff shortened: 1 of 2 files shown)


JustinBeckwith:
LGTM

bshaffer:
Is this going to be necessary with every minor version update? We should consider automating this.

tmatsuo:
Re: automation
That's great. Any ideas for automation are welcome :)

tmatsuo:
Ouch, the e2e tests succeeded but it should fail because of the gcloud error.

tmatsuo:
Investigating

tmatsuo:
I can not reproduce this. Only explanation is that the previous run of the test was canceled, or it failed to delete the version and the test accessed to that version. Maybe I should add unique string to the version number.

tmatsuo:
Tests are truly green (finally). I'm going to merge this. Please let me know if you have any concerns.

PR #36, opened 2015-12-18
Adding some extensions including mongodb.

### LINKED ISSUE #31: Install mongodb extension

tmatsuo:
Fixes #31

### DIFF
--- a/php-nginx/build-scripts/build_php56.sh
+++ b/php-nginx/build-scripts/build_php56.sh
@@ -18,44 +18,42 @@
 # A shell script for installing PHP 5.6.
 set -xe
 
+PHP_SRC=/usr/src/php
+
 curl -SL "http://php.net/get/php-$PHP56_VERSION.tar.gz/from/this/mirror" -o php.tar.gz
 curl -SL "http://us2.php.net/get/php-$PHP56_VERSION.tar.gz.asc/from/this/mirror" -o php.tar.gz.asc
 gpg --verify php.tar.gz.asc
-mkdir -p /usr/src/php
-tar -zxf php.tar.gz -C /usr/src/php --strip-components=1
+mkdir -p ${PHP_SRC}
+tar -zxf php.tar.gz -C ${PHP_SRC} --strip-components=1
 rm php.tar.gz
 rm php.tar.gz.asc
 
-mkdir -p /usr/src/php/ext/memcache
-curl -SL "http://pecl.php.net/get/memcache" -o memcache.tar.gz
-tar -zxf memcache.tar.gz -C /usr/src/php/ext/memcache --strip-components=1
-rm memcache.tar.gz
-
-mkdir -p /usr/src/php/ext/memcached
+mkdir -p ${PHP_SRC}/ext/memcached
 curl -SL "http://pecl.php.net/get/memcached" -o memcached.tar.gz
-tar -zxf memcached.tar.gz -C /usr/src/php/ext/memcached --strip-components=1
+tar -zxf memcached.tar.gz -C ${PHP_SRC}/ext/memcached --strip-components=1
 rm memcached.tar.gz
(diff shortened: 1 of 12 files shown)


php-nginx/build-scripts/build_php70.sh
```diff
@@ -18,20 +18,42 @@
 # A shell script for installing PHP 7.0.
```
sjlangley: why is there a separate script for 70? this means another docker image right?

tmatsuo: For faster development. We will combine them again when we want to go GA.

tmatsuo:
@sjlangley Ping.
OK to merge?

bshaffer:
LGTM - looks like the  `mongo` extension via PECL is not compatible with PHP 7 yet?

tmatsuo:
@bshaffer Thanks. It seems like that [the newest stable release of mongodb](https://pecl.php.net/package/mongodb/1.1.1) supports PHP 7.0. Let me try that before merging.

tmatsuo:
Alright, I can compile the mongodb from pecl now (I couldn't when this PR was created :)

@sjlangley @bshaffer Please take another look.

php-nginx/build-scripts/build_php56.sh
```diff
@@ -118,6 +115,10 @@ strip ${PHP56_DIR}/bin/php ${PHP56_DIR}/sbin/php-fpm
 # Defaults to PHP5.6
 ln -s ${PHP56_DIR} ${PHP_DIR}
 
+# Install shared extensions
+${PHP56_DIR}/bin/pecl install memcache
```
sjlangley: why is this preferable to compiling it above?

tmatsuo: The shell script becomes simpler. Also once if we're sure that the `pecl` command is working fine, it will open a door for us to install other extensions on demand.

For example, possibly we can do that in `composer.sh`, reading `ext-foobar` type of requirements, then install the apt-dev deps, then run `pecl` command.

sjlangley:
lgtm

tmatsuo:
@sjlangley Thanks. I'll squash and merge it.

php-nginx/build-scripts/build_php70.sh
```diff
@@ -82,5 +104,8 @@ make -j"$(nproc)"
 make install
 make clean
 popd
-rm -rf /usr/src/php7
+rm -rf ${PHP_SRC}
 strip ${PHP7_DIR}/bin/php ${PHP7_DIR}/sbin/php-fpm
+
+# Install shared extensions
```
bshaffer: should we install `memcache` for php7 as well?

tmatsuo: Yeah, https://github.com/websupport-sk/pecl-memcache has PHP 7 port.

bshaffer:
I would like to see the extensions get tested in the test suite

tmatsuo:
@bshaffer Good point. Let me add simple tests for them.

tmatsuo:
When adding the tests, I just found that apcu extension was not compiled with PHP 7.0.1. Kudos to @bshaffer

tmatsuo:
@bshaffer I just added the tests. Can you take another look?

bshaffer:
This looks great! The only request I have is that it's easier to see PHP5 has `ereg` and `suhosin`, while PHP7 does not (which I assume is intentional). i.e., define the shared extensions in one place, and then the PHP5 test adds the two specific ones explicitly.

bshaffer:
Nice! LGTM

tmatsuo:
Thanks, I will squash the commits and merge when the test passes.

tmatsuo:
Merging

PR #38, opened 2015-12-22
[TASK] Add https to fastcgi_params

### LINKED ISSUE #37: Nginx doesn't detect the upstream https connection
AppEngines uses HTTPS almost exclusively so it's a good idea to pin HTTPS
to allow applications proper connection detection based on the current protocol.

cedricziel:
AppEngine uses HTTPS per default, but nginx isn't
able to detect it by itself.

Closes #37

### DIFF
--- a/php-nginx/fastcgi_params
+++ b/php-nginx/fastcgi_params
@@ -35,7 +35,8 @@ fastcgi_param   SERVER_ADDR             $server_addr;
 fastcgi_param   SERVER_PORT             $server_port;
 fastcgi_param   SERVER_NAME             $server_name;
 
-fastcgi_param   HTTPS                   $https;
+fastcgi_param   HTTPS                   on;
+fastcgi_param   HTTP_SCHEME             https;
 
 # PHP only, required if PHP was built with --enable-force-cgi-redirect
 fastcgi_param   REDIRECT_STATUS         200;


JustinBeckwith:
So I'm a little curious about what this enables.  App Engine should terminate the HTTPS connection, and then forward requests onto the application container via HTTP.  Why do we need to configure nginx in the app container to handle HTTPS? Calling @tmatsuo and @marslangoog

cedricziel:
This doesnt make nginx handle HTTPS connections-those are handled by the platform as you mentionded, it tells the fastcgi process the correct connection protocol.

The interpreter than acts as if the local webserver was called with https and is able to generate URLs accordingly.

cedricziel:
Let me explain it  in more detail:

AppEngine terminates SSL connections, but uses HTTP connections internally. Because of that, the container-local nginx has no clue about the original HTTPS call. (Come on, AppEngine. It's 2015 and there's the `X-Forwarded-Proto` header. :) ).

Now PHP has no clue about the original incoming HTTPS as well. To make it think we're on HTTPS, the fastcgi params are now hardcoded.

The real problem is the software running on the interpreter which often generates protocol-relative absolute URLs. (https while on https, http while on http.) Which is great, because you can use the dev_appserver.py script and be more flexible in general. To use this without additional modifications, the platform needs to signal the original https context.

I hope I was explicit enough why this is needed - if you need any additional information, please let me know! :)

tmatsuo:
@cedricziel I understand your rationale here, but always setting https=on doesn't seem right to me. Let me see what I can do.

tmatsuo:
There is `HTTP_X_APPENGINE_HTTPS` header which may be `on` or `off` depending on the original request. Does it make sense to set `HTTPS` to this value?

cedricziel:
I cannot answer your question sufficiently, as the header isn't documented and the only occurence is in a python project in the same github org.

If it's a standard header we can use it in the nginx standard config to set https there (which basically would be the same as i did now).

I would really like AppEngine to follow conventions instead and send the conventional proxy headers `X-Forwarded-*`. In the meantime this patch is needed and the image is close to useless without if you generate any absolute URL from inside the application relying on the inbound protocol.

JustinBeckwith:
That's a fair request.  Can we move discussion for this over here?
https://code.google.com/p/googleappengine/issues/detail?id=12252

I'd like to fix this at the platform level if possible, so we don't have to jump through hoops for each runtime.  Let me poke around and see how easy/feasible this is to fix.

tmatsuo:
Once X_FORWARDED_* headers are available, we can just use those values in `nginx.conf` to set `HTTPS` and `HTTP_SCHEME`. We can leverage [the map module] (http://nginx.org/en/docs/http/ngx_http_map_module.html). @JustinBeckwith Or even better, are you planning to set those values at the platform level?

JustinBeckwith:
> Or even better, are you planning to set those values at the platform level?

Not quite sure what you mean here.  Are `HTTPS` and `HTTP_SCHEME` just being set as HTTP headers?

tmatsuo:
@JustinBeckwith Yes. Even if there are X_FORWARDED_* headers available, we need to set `HTTPS` and `HTTP_SCHEME` HTTP headers, which is done in our main nginx configuration.

Since these are language agnostic, it is great if the platform automatically set them.

tmatsuo:
Oh nevermind. `HTTPS` and `HTTP_SCHEME` doesn't seem HTTP headers.

cedricziel:
I just chimed in on the platform ticket.

For everybody else experiencing an issue with incorrect protocols, it's easy to `ADD` a custom `fastcgi_params` file with the changes from this pull request in your custom `Dockerfile`.

```Dockerfile
ADD fastcgi_params /usr/local/nginx/conf/
```

cedricziel:
I'm a bit worried about the fact that the issue is still not resolved. Managed VMs selling point is that I can put a custom runtime in a container. But if the runtime doesnt get the information it needs (and would get on any other platforms that send those headers) - that destroys a lot of value and is a bummer for starters IMHO.

tmatsuo:
@cedricziel I agree. I raised an internal question about it. Let's see how it goes.

BTW, here are relevant RFCs:

http://tools.ietf.org/html/rfc6648
http://tools.ietf.org/html/rfc7239

The former deprecates `X-*` headers, the latter defines `Forwarded` header.

cedricziel:
Alright - thx a bunch for the heads up and the links to the specification! :)

JustinBeckwith:
This is absolutely on the Managed VMs backlog, we just haven't got there yet.

cedricziel:
The issue has been aknowledged on the other tracker. https://code.google.com/p/googleappengine/issues/detail?id=12252

I'll leave this open until it's resolved.

tmatsuo:
Closing in favor of #188

PR #39, opened 2016-01-06
adds no-ansi to prevent verbose composer output

bshaffer:
The following verbose composer output goes on for quite a while. The `--no-ansi` flag prevents this.

screen shot 2016-01-06 at 12 36 41 pm

### DIFF
--- a/php-nginx/composer.sh
+++ b/php-nginx/composer.sh
@@ -81,5 +81,7 @@ EOF
         --no-dev \
         --prefer-dist \
         --optimize-autoloader \
-        --no-interaction
+        --no-interaction \
+        --no-ansi \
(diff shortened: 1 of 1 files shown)


cedricziel:
This actually only removes the color coding. I would prefer that you add `--no-progress` as well.

tmatsuo:
The test is failing because the latest mailparse extension doesn't compile with PHP 5.6.

It will be fixed by #36

bshaffer:
@cedricziel good call!

tmatsuo:
#36 is merged. I just re-run the travis build.

tmatsuo:
The change itself LGTM

PR #41, opened 2016-01-07
Changed the name of php error log

bshaffer:


tmatsuo:
FYI I stopped the travis build since it will just fail

tmatsuo:
#36 will fix the travis build, so after merging #36, please rebase to the HEAD and upload it with `-f` option.

bshaffer:
ok

bshaffer:
although I'm 99% sure the travis PR check will run the merged version, so it should pass w/o the rebase.  The "branch" check (if I pushed this to `google` instead of `bshaffer`) would fail though.

tmatsuo:
Ah, ok, then let's try re-running the test first

php-nginx/php.ini
```diff
@@ -586,7 +586,7 @@ html_errors = Off
 ; Example:
 ;error_log = php_errors.log
 ; Log errors to syslog (Event Log on Windows).
-error_log = /var/log/app_engine/php-error.log
+error_log = /var/log/app_engine/application.log
```
tmatsuo: What this change will give us? Will this give you a better name in the dropdown in the Developers Console?

php-nginx/php.ini
```diff
@@ -478,7 +478,7 @@ error_reporting = E_ALL & ~E_DEPRECATED & ~E_STRICT
 ; Development Value: On
 ; Production Value: Off
 ; http://php.net/display-errors
-display_errors = Off
+display_errors = stderr
```
tmatsuo: According to http://php.net/manual/en/errorfunc.configuration.php#ini.display-errors, `display_errors` should not be used on production systems.

Is it really OK to turn it on?

tmatsuo: FYI Travis build is failing because phpinfo.php shows a warning like:
```
Warning: phpinfo() has been disabled for security reasons in /app/phpinfo.php on line 19
```

bshaffer: Hmm, I see your point. Unfortunately, this is the only way (I know of) to
separate stdout and stderr in Php. That warning is probably meant for when
display_errors is set to On.

Also, I'd be fine with setting log_errors to On and combining them. So it
depends on how important we think it is to separate them. Personally I'd
like to try this out and see if it works as expected.

tmatsuo: When I tested these configs, the logs sent to STDERR don't appear anywhere. Also, it is possible to override any `php.ini` settings by having `php.ini` file at the top level directory of your web app. I would not change the default values to something less strict.

Good news is that I think I found an alternate solution.

When I set `catch_workers_output = yes` in `/usr/local/php/etc/php-fpm.conf`, I see the logs in `/var/log/app_engine/php-fpm.log`. 

Does it work for you?

php-nginx/php.ini
bshaffer: This allows the fluentd logger to detect the log file. Currently, it is not capturing the log output because it doesn't match the [fluentd configuration](https://github.com/GoogleCloudPlatform/appengine-sidecars-docker/blob/master/fluentd_logger/managed_vms.conf#L4)

php-nginx/php.ini
bshaffer: Yes, but if `error_log` is set, it writes there also. I don't think we care to write two files with the same logs. Especially if we set `log_errors` to `On`

php-nginx/php.ini
tmatsuo: OK. However, I don't see anything in that file (php-error.log or application.log) when I emit text to stdout/stderr.

php-nginx/php.ini
tmatsuo: I'm puzzled. In my experiment, I don't see anything logged in a file specified by `error_log` in `php.ini`.

php-nginx/php.ini
tmatsuo: How about `app-error.log`?

PR #42, opened 2016-01-07
Changed the log file names so that our fluentd picks them.

tmatsuo:
Also let php-fpm catches the stdout and stderr and stop picking up
php-fpm's access log.

Background: [The current fluentd configuration](https://github.com/GoogleCloudPlatform/appengine-sidecars-docker/blob/master/fluentd_logger/managed_vms.conf) only picks up certain file name patterns. We need to adapt our log file names to their pattern.

### DIFF
--- a/php-nginx/php-fpm.conf
+++ b/php-nginx/php-fpm.conf
@@ -44,7 +44,7 @@ pid = /var/run/php-fpm.pid
 ; in a local file.
 ; Note: the default prefix is /usr/local/php/var
 ; Default Value: log/php-fpm.log
-error_log = /var/log/app_engine/php-fpm.log
+error_log = /var/log/app_engine/app-php-fpm.log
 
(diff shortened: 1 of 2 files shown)


bshaffer:
LGTM

php-nginx/php-fpm.conf
```diff
@@ -388,7 +388,7 @@ pm.max_spare_servers = 3
 
 ; The access log file
 ; Default: not set
-access.log = /var/log/app_engine/$pool.access.log
+access.log = /var/log/app_engine/php-fpm-$pool.access.log
```
tmatsuo: Stop picking up this log entry in the `app` category.

tmatsuo:
I'll merge this after the test passes.

tmatsuo:
The new image will be uploaded shortly :)

PR #43, opened 2016-01-08
Install mercurial and subversion for composer.

### LINKED ISSUE #26: Additional VCS packages needed
This is a followup issue to #18 .

Composer supports several different VCS drivers (listed [here](https://github.com/composer/composer/tree/master/src/Composer/Repository/Vcs)).

Git was introduced in #18, but the remaining VCS packages should be included into the images as well.

tmatsuo:
Fixes #26

### DIFF
--- a/php-nginx/Dockerfile
+++ b/php-nginx/Dockerfile
@@ -36,6 +36,8 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
     libxml2 \
     libxslt1.1 \
     logrotate \
+    mercurial \
+    subversion \
     supervisor \
     zlib1g
 


tmatsuo:
@bshaffer @JustinBeckwith Give me a quick LGTM

JustinBeckwith:
LGTM.  One little thing - for Ruby, we keep a table that lets us remember why we install the things we're installing:
https://github.com/GoogleCloudPlatform/ruby-docker/blob/master/appengine/Dockerfile#L5

May be nice to do that here too.  That way in a year we don't come back and say "why would we have subversion, lets rip that out" :)

tmatsuo:
Thanks. You can refer to the commit message then you'll see why :)

PR #46, opened 2016-01-25
Update PHP versions to 7.0.2 and 5.6.17

cedricziel:
Those include security updates and date back to January 07.

Do you plan to implement a process to keep track of such events?


tmatsuo:
@cedricziel Thanks! I agree that we should have a process, ideally automated, but manual one is OK to start. Somehow the travis build is failing, so I restarted and see how it goes.

tmatsuo:
Tests all green. Merging

PR #47, opened 2016-01-29
Set environment variables for memcache based session.

tmatsuo:
Based on a discussion on https://github.com/cakephp/app/issues/331


cedricziel:
It's the right approach for a better out-of-the-box experience, I think.

tmatsuo:
@cedricziel 

According to the further discussion, it turned out that you can use environment variables in php.ini.
I open #48 which adding an e2e test showing how to enable memcached based session.

PR #48, opened 2016-01-29
Enable memcached session automatically on prod.

tmatsuo:
This PR automatically enable memcached session on prod.

@bshaffer 
@sjlangley 
@cedricziel 

Please take a look and let me know what you think.

### DIFF
--- a/php-nginx/build-scripts/build_php70.sh
+++ b/php-nginx/build-scripts/build_php70.sh
@@ -54,7 +54,7 @@ rm -f configure
 ./buildconf --force
 ./configure \
     --prefix=$PHP7_DIR \
-    --with-config-file-scan-dir=$APP_DIR \
+    --with-config-file-scan-dir=$APP_DIR:${PHP7_DIR}/lib/conf.d \
     --disable-cgi \
     --disable-memcached-sasl \
     --enable-apcu \
@@ -104,6 +104,9 @@ make clean
 popd
 rm -rf ${PHP_SRC}
 strip ${PHP7_DIR}/bin/php ${PHP7_DIR}/sbin/php-fpm
+
+# Create a directory for additional config files.
+mkdir -p ${PHP7_DIR}/lib/conf.d
 
 # Install shared extensions
 ${PHP7_DIR}/bin/pecl install mongodb
--- a/php-nginx/entrypoint.sh
+++ b/php-nginx/entrypoint.sh
@@ -38,6 +38,14 @@ if [ -f ${APP_NGINX_CONF} ]; then
     mv ${APP_NGINX_CONF} ${NGINX_DIR}
 fi
 
+# Configure memcached based session.
+if [ -n ${MEMCACHE_PORT_11211_TCP_ADDR} ] && [ -n ${MEMCACHE_PORT_11211_TCP_PORT} ]; then
+    cat <<EOF > ${PHP_DIR}/lib/conf.d/memcached-session.ini
+session.save_handler=memcached
+session.save_path="${MEMCACHE_PORT_11211_TCP_ADDR}:${MEMCACHE_PORT_11211_TCP_PORT}"
+EOF
+fi
+
 # Configure document root in php.ini and nginx.conf with DOCUMENT_ROOT
 # environment variable or APP_DIR if DOCUMENT_ROOT is not set.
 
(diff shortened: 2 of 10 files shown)


cedricziel:
IMHO this should go into the documentation and not into the main php.ini.

It's quite cool to be able to use memcached through some environment variables, but I vote for less magic and more documentation. :)

I'm fine with the change though :+1:

tmatsuo:
@cedricziel Thanks for the feedback! I added the documentation for this.

cedricziel:
Awesome! Thx :)

tmatsuo:
@raggi Can you take a look?

raggi:
I would actually strongly vote in the other direction, and go for convention over configuration. Users should be able to get more things going with less reading and less configuration than forcing them to manually read and configure lots of features that are configured as-standard in competing environments.

As far as the impact on using the image in GKE and so on, this is a secondary goal, not a primary goal. Our primary goal is seamless UX for GAE users.

GKE users by nature of GKE know they'll have to do further configuration. If their apps don't use these features, the env vars being missing won't affect them. If they do, it's trivial to add the env vars in your pod configuration. It's actually harder work for the user to customize their app image to change php.ini and keep it up to date over time.

tmatsuo:
@raggi 

Alright. Do you prefer having those configurations by default?

raggi:
Other than the discussion above, the code looks fine.

Am I correct in assuming that the current default session storage behavior is files, and that's actually broken for more than one node?

tmatsuo:
@raggi 

> Other than the discussion above, the code looks good.

Thanks. Let me make these configurations by default.

> Am I correct in assuming that the current default session storage behavior is files, and that's actually broken for more than one node?

Yes.

raggi:
Great, thanks, yes, lets absolutely make this work for multinode configurations by default.

cedricziel:
Sounds reasonable.

tmatsuo:
@raggi @cedricziel 

I updated this PR, and now we automatically enable the memcached based session if there are relevant envvars.

PTAL

php-nginx/build-scripts/build_php70.sh
```diff
@@ -105,6 +105,9 @@ popd
 rm -rf ${PHP_SRC}
 strip ${PHP7_DIR}/bin/php ${PHP7_DIR}/sbin/php-fpm
 
+# Create a directory for additional config files.
+mkdir ${PHP7_DIR}/lib/conf.d
```
raggi: add a -p flag, just to save any issues

php-nginx/entrypoint.sh
```diff
@@ -38,6 +38,14 @@ if [ -f ${APP_NGINX_CONF} ]; then
     mv ${APP_NGINX_CONF} ${NGINX_DIR}
 fi
 
+# Configure memcached based session.
+if [ -n ${MEMCACHE_PORT_11211_TCP_ADDR} ] && [ -n ${MEMCACHE_PORT_11211_TCP_PORT} ]; then
```
raggi: I would write this:
```if [[ -n "${MEMCACHE_PORT_11211_TCP_ADDR}${MEMCACHE_PORT_11211_TCP_PORT}" ]];
then```

**Comment by raggi** [comment]
Couple of minors :+1:

File: php-nginx/build-scripts/build_php70.sh (hunk 1 above)
Comment (tmatsuo) [inline php-nginx/build-scripts/build_php70.sh]: Good call, done.

File: php-nginx/entrypoint.sh (hunk 2 above)
Comment (tmatsuo) [inline php-nginx/entrypoint.sh]: By checking both, it will prevent unintentional failure when either of them is missing. If we only check the combined string, there will be a chance of having incomplete configuration.

Or does it accept the save path like these?

```
; is it localhost?
session.save_path=:11211
```
or
```
; Is it the default port?
session.save_path=172.172.172.172: 
```

raggi: That's fair, although I'm not sure if it's worse to have the extension cause an error or to have it silently fall back to files mode.

Keep them split is fine.

tmatsuo:
I squashed the commits. I will merge this after the test passes.

tmatsuo:
I found that I made a mistake for checking the return value from the gcloud deployment, so I fixed it.
I will merge this after the test pass.

tmatsuo:
Alright, merging. Thanks @raggi @cedricziel

PR #53, opened 2016-02-08
Prefers 5.6 if both 5.6 and 7.0 satisfies the req.

### LINKED ISSUE #52: Prefer PHP 5.6 over PHP 7.0
It seems like the memcached extension is not stable on PHP 7.0.

https://github.com/php-memcached-dev/php-memcached/issues/213

We should prefer PHP 5.6 until it's stable enough.

tmatsuo:
Fixes #52


scripts/build_images.sh
```diff
@@ -19,4 +19,5 @@ docker build -t php-nginx php-nginx
 docker build -t php56 testapps/php56
 docker build -t php56_custom testapps/php56_custom
 docker build -t php70 testapps/php70
+docker build -t php56_70 testapps/php56_70
```
bshaffer: This is a build with both PHP versions installed

bshaffer:
LGTM, although this sure looks like a lot of images to build. Is that standard?

scripts/build_images.sh
tmatsuo: This is a virtual user app, which has a composer entry of php "^5.6|^7.0". For now we should choose PHP 5.6 for such apps.

tmatsuo:
The images are for testing and I think it's fair.

tmatsuo:
Thanks, merging!

PR #55, opened 2016-02-17
Increased the memory limit for composer.

tmatsuo:


tmatsuo:
Test failed due a gcloud sdk change

tmatsuo:
Oh, that's not true. This is a path issue.

tmatsuo:
That's not true too!

The gcloud installation failed
```
ERROR: (gcloud.components.update) The component [bq] failed to download.
```

tmatsuo:
I will just retry the build

bshaffer:
This is a little scary - an unlimited memory limit can cripple the machine if the process consumes all the available memory. Can you link to the problem this is trying to fix?

tmatsuo:
@bshaffer 
There is an internal bug at: b/27207294

but basically the composer fails with:
```
Fatal error: Allowed memory size of 134217728 bytes exhausted (tried to allocate 24 bytes) in phar:///usr/local/bin/composer/src/Composer/DependencyResolver/RuleSe
tGenerator.php on line 126
Check https://getcomposer.org/doc/articles/troubleshooting.md#memory-limit-errors for more info on how to handle out of memory errors.
```

tmatsuo:
Also, this runs only during the build process, it doesn't affect the serving stack at all.

cedricziel:
As for the "scary memory limit" that's something the platform should take of, now that Docker supports memory limits natively.

bshaffer:
Interesting that composer claims to [increase to 1GB internally](https://getcomposer.org/doc/articles/troubleshooting.md#memory-limit-errors), but doesn't seem to be doing that here. I'd like to know if something changed. But in general, LGTM.

bshaffer:
see https://github.com/composer/composer/blob/15face5432d7b7334db6ac69fac0190971cafa6e/bin/composer#L36

bshaffer:
oh, I bet it's because we've blacklisted `ini_set`

bshaffer:
I would be more comfortable setting this to `1GB`. What is the argument for setting to `-1` instead? Do we actually think we'll need more than `1GB` here?

tmatsuo:
@bshaffer Nice find :) It's good to know what's happening here.

There is no reason to limit the memory amount in any sense, because if the composer process fails, the deployment will fail anyways.

tmatsuo:
OK to merge, or do you want me to make it 1GB?

tmatsuo:
Ok, merging. It is a good practice to have a memory limit, but it doesn't make sense in this case.

PR #57, opened 2016-02-19
Added grpc extension.

tmatsuo:
We need to use 0.6.0 for now. Also it's not available for PHP 7.0.

### DIFF
--- a/php-nginx/Dockerfile
+++ b/php-nginx/Dockerfile
@@ -17,12 +17,14 @@
 FROM gcr.io/google_appengine/base
 
 # persistent / runtime deps
+RUN echo "deb http://http.debian.net/debian jessie-backports main" >> /etc/apt/sources.list
 RUN apt-get update && apt-get install -y --no-install-recommends \
     cron \
     curl \
     gettext \
     git \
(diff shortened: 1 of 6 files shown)


tmatsuo:
C'mon Travis, you can do it!

php-nginx/Dockerfile
```diff
@@ -17,12 +17,14 @@
 FROM gcr.io/google_appengine/base
 
 # persistent / runtime deps
+RUN echo "deb http://http.debian.net/debian jessie-backports main" >> /etc/apt/sources.list
```
JustinBeckwith: I would rather not take a dependency on backports. Let's chat offline.

tmatsuo: Ah good point. It may introduce vulnerabilities. I will likely close this PR

tmatsuo: Closing this. We will come back when the backport dep is not necessary.

bshaffer: We can also [compile it](https://github.com/grpc/grpc/blob/master/INSTALL#L25), since who knows when that will happen...

PR #59, opened 2016-02-19
Unset max_input_time when running composer.

### LINKED ISSUE #58: composer fails with "Maximum execution time of 0 seconds exceeded"
https://travis-ci.org/GoogleCloudPlatform/php-docker/builds/110488822#L8865
```
Fatal error: Maximum execution time of 0 seconds exceeded in phar:///usr/local/bin/composer/vendor/symfony/console/Helper/Table.php on line 196
The command '/bin/sh -c /composer.sh' returned a non-zero code: 255
```

We should temporary set max_input_time=0 (the error message is confusing, but the max_input_time is the cause)

tmatsuo:
fixes #58

### DIFF
--- a/php-nginx/composer.sh
+++ b/php-nginx/composer.sh
@@ -77,6 +77,7 @@ EOF
         -d suhosin.executor.func.blacklist=none \
         -d disable_functions= \
         -d memory_limit=-1 \
+        -d max_input_time=-1 \
         /usr/local/bin/composer \
         install \
         --no-dev \


bshaffer:
shouldn't this be `-d max_input_time=-1` to ensure it cannot timeout?

bshaffer:
Also I believe the flag is `max_execution_time` (see [here](http://php.net/manual/en/info.configuration.php#ini.max-execution-time))

tmatsuo:
@bshaffer The error message is a lie. We have max_execution_time=86400 and it shouldn't be the cause. Also 0 and -1 is the same meaning.

bshaffer:
We should be consistent and use `-1` like we are for memory limit.

tmatsuo:
Done!

bshaffer:
thank you! I know it's a nit but I appreciate it.

LGTM!

tmatsuo:
Thanks!

PR #62, opened 2016-03-01
Added grpc extension.

tmatsuo:


JustinBeckwith:
Well that didn't take long!  LGTM

tmatsuo:
Yeah, I actually haven't build the docker image locally, so it is possible to fail building. Let's see the travis result.

tmatsuo:
Seems like we still need `-beta`.

bshaffer:
hey! nice. Is the `pecl` command sufficient to install all gRPC's dependencies?

bshaffer:
i.e. maybe we should include a gRPC call as part of the tests. To the URL shortener, for example.

tmatsuo:
Sure, it is nicer to have such tests. I'll try to add e2e tests accessing Cloud Pub/Sub (I don't think URL shortner has enabled gRPC).

tmatsuo:
I added an end to end test. Let's see the travis build result.

testapps/php56_e2e/composer.json
```diff
@@ -1,6 +1,15 @@
 {
+    "repositories": [
+        {
+            "type": "vcs",
+            "url": "https://github.com/stanley-cheung/Protobuf-PHP"
```
bshaffer: Is @stanley-cheung interested in submitting his changes `datto/protobuf-php`? Why are we using his fork?

testapps/php56_e2e/web/grpc_pubsub.php
```diff
@@ -0,0 +1,44 @@
+<?php
+
+require_once __DIR__ . '/vendor/autoload.php';
+require_once 'empty.php';
+require_once 'pubsub.php';
+require_once 'timestamp.php';
+require_once 'descriptor.php';
+require_once 'http.php';
+require_once 'annotations.php';
```
bshaffer: We could add these to the autoloader with composer's [classmap directive](https://getcomposer.org/doc/04-schema.md#classmap) in [`testapps/php56_e2e/composer.json`](https://github.com/GoogleCloudPlatform/php-docker/blob/master/testapps/php56_e2e/composer.json):

```php
{
    ...
    "autoload": {
        "classmap": ["web"]
    }
}
```

tmatsuo:
Stopped the travis build for merging other PR quicker

testapps/php56_e2e/composer.json
stanley-cheung: We tried, but the upstream repo has not responded in a year: https://github.com/drslump/Protobuf-PHP/pull/48

But because he has submited his package to Packagist, we can't submit our fork unless we change the repo name.

That was why I have to resort to this `composer.json` syntax to use his official package on Packagist in the `require` section but substitute the actual content with our fork.

testapps/php56_e2e/web/grpc_pubsub.php
tmatsuo: done

testapps/php56_e2e/composer.json
bshaffer: Shouldn't this get pulled in implicitly as part of the [composer requirements](https://github.com/grpc/grpc/blob/master/src/php/composer.json#L2) for grpc?

stanley-cheung: That is true. By pulling `grpc/grpc`, you are also implicitly pulling `datto/protobuf-php` (with a substituted fork) yes.

tmatsuo: Ok, so there's no need for specifying datto/protofuf-php here. I will remove it

stanley-cheung: Sorry I made a mistake. See my update above.

tmatsuo: No problem, I added it back.

tmatsuo:
Oooh, of course the auto generated PHP files don't pass our coding style check...

testapps/php56_e2e/source-contexts.json
```diff
@@ -0,0 +1 @@
+[{"labels": {"category": "remote_repo"}, "context": {"git": {"url": "git@github.com:GoogleCloudPlatform/php-docker.git", "revisionId": "9abd7f7f4adeaf400c774ed667553eec74a385c4"}}}, {"labels": {"category": "remote_repo"}, "context": {"git": {"url": "git@github.com:tmatsuo/php-docker.git", "revisionId": "9abd7f7f4adeaf400c774ed667553eec74a385c4"}}}]
```
tmatsuo: Hey who are you man!

tmatsuo:
I fixed several things, but the test still fails:

```
2016/03/04 02:29:40 [error] 321#0: *11 FastCGI sent in stderr: "PHP message: PHP Fatal error:  Call to a member function getTopics() on null in /app/web/grpc_pubsub.php on line 36" while reading response header from upstream, client: 172.17.0.5, server: , request: "GET /grpc_pubsub.php HTTP/1.1", upstream: "fastcgi://unix:/var/run/php5-fpm.sock:", host: "tmp-dot-php-mvm-a.appspot.com"
```

Seems like `ApplicationDefaultCredentials::getCredentials` is not working. I'll investigate tomorrow.

tmatsuo:
I think I found a cause!!!

```
Invalid resource name given (name=projects/)
```

tmatsuo:
@bshaffer The test is green now. PTAL

testapps/php56_e2e/composer.json
bshaffer: OK gotcha. I advocate making it the default, since it will be very difficult for users to discover the need for them to override with your fork on their own.

testapps/php56_e2e/web/grpc_pubsub.php
```diff
@@ -0,0 +1,38 @@
...
+
+$deadline = Grpc\Timeval::InfFuture();
+
+$req = new google\pubsub\v1\ListTopicsRequest();
+$proj = getenv('GAE_LONG_APP_ID');
+$req->setProject('projects/' . $proj);
+$call = $client->ListTopics(
+    $req,
+    [],
+    ['call_credentials_callback' => 'updateAuthMetadataCallback']);
```
bshaffer: It would be nice if we could use the credentials' [`getUpdateMetadataFunc`](https://github.com/google/google-auth-library-php/blob/master/src/CredentialsLoader.php#L140) function for this:

```php
$auth_credentials = ApplicationDefaultCredentials::getCredentials(
     ['https://www.googleapis.com/auth/pubsub']
);
$call = $client->ListTopics(
     $req,
     [],
     ['call_credentials_callback' => $auth_credentials->getUpdateMetadataFunc())];
```

But it looks like since the first argument doesn't match up, this won't work. Maybe we should change the `google/auth` library to handle this? I assume that is what this function is meant for.

bshaffer:
LGTM

testapps/php56_e2e/web/grpc_pubsub.php
stanley-cheung: hm, so this callback `'updateAuthMetadataCallback'` is meant to be seen in gRPC's perspective. It's as if the gRPC core library is saying, "I am ready for the auth credentials. Here's what we know about this call so far, contained in the `$context` variable (the most important one being `$context->service_url`, coz after load balancing, user may not know which is the host until now)". So with this `$context` variable, we ask the auth library for the auth header by calling `$auth_credentials->updateMetadata` passing `$context->service_url` to it. 

To me these 2 callbacks don't need to lineup because the outer gRPC callback can do more than just Auth. I think it;s OK to leave this as is.

testapps/php56_e2e/composer.json
stanley-cheung: So I just tried this again yesterday. Even if the root level `grpc/composer.json` is going to pull in `datto/protobuf-php` and with my `stanley-cheung` override, as a user of `grpc/grpc`, you are not going to get the `stanley-cheung` override unless the user also puts that in their own `composer.json`. In other words, as a user of `grpc/grpc`, at the minimum, they still need to `require grpc/grpc` and add the `stanley-cheung/protobuf-php` override. It's not ideal but I can spell it out more in grpc's documentation.

testapps/php56_e2e/web/grpc_pubsub.php
bshaffer: Ok, thanks for the explanation. Is it a problem that the `google/auth` library isn't doing anything with the `service_url`?

Also, we could make this cleaner like this: 

```php
$auth_credentials = ApplicationDefaultCredentials::getCredentials(
     ['https://www.googleapis.com/auth/pubsub']
);
$call = $client->ListTopics(
     $req,
     [],
     ['call_credentials_callback' => function($context) use ($auth_credentials) {
         return $auth_credentials->updateMetadata([], $context->service_url);
     })];
```

stanley-cheung: I believe the $service_url is used only in some xxxCredentials class, like `ServiceAccountJwtAccessCredentials` but not the others.

I like the inline function change.

bshaffer: ahh yes, I see that now. Very cool. SGTM, thanks @stanley-cheung!

tmatsuo: Done

tmatsuo:
I just stopped the build, I will re-run later today.

tmatsuo:
Alright, merging.

testapps/php56_e2e/composer.json
bshaffer: @stanley-cheung Yes, we should add this to composer. Also, are you sure the override is required in the userland `composer.json`? If so, that seems like a bug in composer. And also if so, we could submit your fork as a new composer package, which would definitely fix the issue.

Requiring the users to override in `composer.json` rather than just call `composer require grpc/grpc` is not acceptable DX IMHO. We could fork to `GoogleCloudPlatform` and submit a new package from there.

stanley-cheung: Yes I have tried this again and an end-user needs to do this override, and I agree that it's non-ideal. OTOH, this was supposed to be a temporary bridgegap before we have official PHP support in google/protobuf - it's being implemented now. So if we can wait for a few months, this will all go away and we will have the official solution.

PR #66, opened 2016-03-08
Install composer dev-dependencies on build

cedricziel:
To be able to use environment configuration in composer scripts
like for example Symfony 2 does, it is necessary to either defer
the execution of composer scripts up to a point where the developer
can set custom runtime configuration (through ENV in the Dockerfile),
which would harm the onboarding-experience as a custom Dockerfile is
necessary at all times to execute scripts, or install dev-depenencies
as well in the first place, to have them available when composer
scripts need them, because they don't know which environment they
run in.

This change applies the latter: Optimizing for production by removing
the dev-dependencies is now up to the inheriting image.

Dockerfile:
```
FROM gcr.io/php-mvm-a/php-nginx:latest

ENV SYMFONY_ENV prod

RUN cd $APP_DIR && $PHP_DIR/bin/php \
    -d suhosin.executor.include.whitelist=phar \
    -d disable_functions= \
    -d memory_limit=-1 \
    -d max_input_time=-1 \
    /usr/local/bin/composer \
    install \
    --prefer-dist \
    --no-dev \
    --optimize-autoloader \
    --no-interaction \
    --no-ansi \
    --no-progress
```

This is necessary, because environment variables defined in the inheriting Dockerfile or app.yaml are not exposed to the container while it's being build.

### DIFF
--- a/php-nginx/composer.sh
+++ b/php-nginx/composer.sh
@@ -80,7 +80,6 @@ EOF
         -d max_input_time=-1 \
         /usr/local/bin/composer \
         install \
-        --no-dev \
         --prefer-dist \
         --optimize-autoloader \
         --no-interaction \


tmatsuo:
@cedricziel 

I only see the one line removal change in the Dockerfile in this pull request. I think you intended to add https://github.com/cedricziel/php-docker/commit/90551e92f40b15432b315c827389e29c07b8e0f5 to this pull request.

tmatsuo:
Also why do you need the `dev` dependencies on production?

cedricziel:
That's the whole point of this PR - We don't know since we cannot, to my knowledge, set environment in the build environment. To keep an application portable,  it needs the dev dependencies as well because in many cases, like with symfony, it assumes the 'dev' environment if nothing is set.

IMHO it's a small trade-off loading the dev dependencies by default compared to the users needing to modify their application upfront. Sorry to call a competitor again-but they are pretty good at keeping the entry barriers low. - Optimizations should need to be done manually if the user wants them, when they harm the on-boarding experience if they're applied automatically. 

Yeah, good point about the second commit - I planned to place an additional PR to show the problem + the solution afterwards.

cedricziel:
I'll push to example branches when I've ran the scripts locally to proof my point :)

Well.. The more I think about it.. Maybe we should just not run composer scripts in this image, but just install and defer the scripts to the inheriting image. That might result in a much better experience.

Heroku does the following in their docker images: 
* Install *everything* from the composer dependencies to give the buildpack complete freedom and as a result of this, the user can influence the environment on a later stage (such as a userscript when the container starts, or a buildpack in their case https://github.com/heroku/docker-php/blob/master/Dockerfile#L74).
* in the buildpack they run `composer install --no-dev --prefer-dist --optimize-autoloader --no-interaction` at a stage that is *after* the initial image has been baked, so user environment can kick in. (https://github.com/heroku/heroku-buildpack-php/blob/master/bin/compile#L237)

In the GCP php-docker image we cannot do all this automatically, because the docker image build is a sort of one-off job which has exactly one stage (as far as we can condense this :) ): The Docker build on deploy to the docker builder. That's why we need to prepare the image for at least the flexibility to correct anything  that's been done in the base `php-docker` image in an inheriting image. If we don't install all dependencies and forbid to influence what scripts should be run on `composer.json`, then the flexibility is gone and as a developer I have to modify my application again to make it run on managed vms, even if they follow the 12 factor style.

tmatsuo:
Sorry, I still don't understand the original problem is.

With the `--dev` option, what exact problem you see?

cedricziel:
Actually it's the `--no-dev` option that makes assumptions it cannot verify at a place in time the user cannot influence anything. Look at the symfony example: The composer install scripts kick in with the assumption they're in the dev environment (#67) and fail the whole build of the image (please take a look at the travis output to see why it fails).

As stated above, alternatively the `--no-scripts` option could be used instead of `--no-dev`.

All this comes down to the one point: The assumption that an application put in the container does rely only on prod-dependencies at any point in time is purely academical and a restriction that's close to the original AppEngine sandbox-which managed vms try to overcome with the promise of greater flexibility.

**BUT:** The moment the image makes assumptions that the user of the image cannot overcome it's useless since other PaaS's allow for flexibility in that point and I'd need to build a custom image from the ground up.

I hope that made things a little clearer and I'd be glad if we could discuss this further if needed.

tmatsuo:
Sorry, but to me, it seems just like a bug in the Symfony2 script or dependency. If that script is needed for production run, the dependency in question should be in the main deps. If that script is just optional, it should not error out on prod.

cedricziel:
Hmm we have problem with docker here - neither with the application nor with the scripts. With docker we have the sequence of build steps which are set in stone - this also involves environment variables. They can only be introduced the moment the sequence reaches the user-provided `Dockerfile`. Symfony on the other hand relies on the environment to be present (`SYMFONY_ENV` to be precise) to determine what environment it's running in. But with the current configuration it's impossible to know or even set the required environment since I can only ever set new environment variables *after* the initial install from `composer.sh` is run-there's 0 chance to build 12 factor apps without being able to configure the environment.

Let me bring this to a good end: I  propose to add the `--no-scripts` flag then instead of removing the `--no-dev` one. That way neither the environment makes assumptions, nor the application is faced with requirements it cannot cope with because it's just too early in the flow. Users that rely on composer hooks can run them in their user-provided `Dockerfile` and only `prod` dependencies get installed.

tmatsuo:
Why does symfony2 assume it's a `dev` environment when there's no environment variable? In my opinion, it should default to prod because it's safer.

Re: Adding `--no-scripts`,
I'm wondering if this will add another headache for other frameworks.

Can we just always add SYMFONY_ENV=PROD in `composer.sh` for a temporary workaround?

cedricziel:
I think the default environment is a matter of taste - yet Symfony shares the default environment being development with rails and spring for example. I find it pretty reasonable.

Yes, adding `--no-scripts` removes all side-effects a `composer install` has. - Symfony normally re-creates the directory structure under the `var` directory, warms the cache and re-writes its configuration (https://github.com/symfony/symfony-standard/blob/master/composer.json#L31). Laravel does the same (https://github.com/laravel/laravel/blob/master/composer.json#L38). So it would be up to the user in an inheriting image to implement the further actions the current application in the container needs. - Which is by far better than forcing anything you cannot escape.

Unfortunately-as I mentioned earlier-currently we only have one install phase (please look at the Heroku image example again - adding all the dependencies without scripts first and then installing again with scripts is great).

Regarding a default environment in the image - I don't think that's a good idea. Why not e2e, test or staging? We really don't know which environment the user sets up.

Give the user some freedom. Maybe it's just the current implementation of the image and it's build scripts that doesnt nail it a 100 %.

tmatsuo:
Ok, now I fully understand the problem. Thanks @cedricziel 

Regardless it's e2e, test, or staging, if the app runs on MVMs, it should default to production environment. `Dev` environments tend to be loose, often reveal developers' secret, sometimes vulnerable. That's why I want to stick with `--no-dev` option by default.

As you suggested, we can add `--no-scripts` in the first stage. I also don't like all the `-d` options when running `composer`, so I'm thinking to provide a shell script wrapper for running composer, so that you can just do like:

```
# Need this env to run the script
ENV SYMFONY_ENV prod
RUN cd $APP_DIR && composer install \
    --prefer-dist \
    --no-dev \
    --optimize-autoloader \
    --no-interaction \
    --no-ansi \
    --no-progress
```

How does it sound?

cedricziel:
Sure - `--no-scripts` is nice and a lot better.

PR #68, opened 2016-03-10
Add Symfony2 test app with dev-dependencies fix applied

cedricziel:
This PR adds a Symfony2 application with almost no configuration (2 code style changes and a different logger) changes.

* adds the application to the tests to check if it can run
* modifies the code style checks to use a different CS (the symfony one which is PSR2 + X) for the
  new example app

Same as #67 with #66 applied + some Dockerfile and supervisord.conf sugar on top.


cedricziel:
It's really up to you how you organize your repositories :)

I'd be glad to have at least one real-world example in the build besides the simple ones.

cedricziel:
I'll close the PR, mainly because of #69 and #66. Still I think you should consider some more sophisticated examples.

PR #69, opened 2016-03-12
Prevent composer from running any scripts on install

cedricziel:
All users that rely on composer scripts will need to extend
the Dockerfile and run the scripts manually in a second
build-step in their Dockerfile step.


tmatsuo:
@cedricziel Thanks!
LGTM, I'll merge once the test passes

Issue #65, opened 2016-03-06
Raise pm.max_children or make it configurable


tmatsuo:
The current 5 is too small

tmatsuo:
or should we make it configurable?

cedricziel:
IMHO it should depend on the CPU size of the current box if any :)

tmatsuo:
@cedricziel also machine memory and your app's memory consumption, so automatic calculation is difficult. For a time being, allowing php-fpm.conf override by having the following line

```
include /app/php-fpm.conf
```

may be a short time solution. Users may put php-fpm.conf override in your project directory.

How does it sound?

PR #70, opened 2016-03-12
Use TCP for php-fpm and increase the number of workers.

tmatsuo:
When using the unix socket, I often saw 502 GATEWAY error. The log shows something like this:
```
2016/03/12 10:25:23 [error] 18#0: *6016 connect() to unix:/var/run/php5-fpm.sock failed (11: Resource temporarily unavailable) while connecting to upstream, client: 172.17.0.6, server: , request: "GET / HTTP/1.1", upstream: "fastcgi://unix:/var/run/php5-fpm.sock:", host: "tmatsuo-wordpress.appspot.com"
```

First I tried to raise net.core.somaxconn with sysctl, but it raises an error like "read only file system". The next attempt is using TCP socket instead of unix socket. With my experiments, TCP sockets are much stabler upon high load.

See also #65

### DIFF
--- a/php-nginx/README.md
+++ b/php-nginx/README.md
@@ -243,6 +243,10 @@ deployment (`app.yaml` or `Dockerfile` are both fine).
 As this is your personal credential for GitHub, please make sure you do not
 commit this token to your repository.
 
+## Reserved TCP port
+
+Nginx and php-fpm are communicating via the TCP port 9000.
+
 ## Deploy with gcloud
 
 You can deploy the app by:
--- a/php-nginx/nginx.conf
+++ b/php-nginx/nginx.conf
@@ -37,6 +37,11 @@ http {
 
     keepalive_timeout  65;
 
+    upstream php-fpm {
+        server 127.0.0.1:9000 max_fails=3 fail_timeout=3s;
+        keepalive 8;
+    }
+
     server {
         listen       8080;
         root %%DOC_ROOT%%;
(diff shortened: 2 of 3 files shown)


tmatsuo:
@cedricziel thanks, changed the port to 9000

tmatsuo:
Seems like a good idea, will do thanks

tmatsuo:
Oh, presumably the test is failing because it couldn't start 10 processes.

tmatsuo:
@bshaffer Can you take a look?

bshaffer:
So we're switching to TCP instead of a unix socket because it's "more stable". Can you explain a little bit what this means?

bshaffer:
Otherwise, LGTM

tmatsuo:
The default value of the kernel parameter `net.core.somaxconn` is 128 and now it's not possible to change it from within the docker image. Upon high load when the server starts to have more than 128 connections, it will throw 502 error.

Generally, unix socket is faster, so once we have a way to change the `net.core.somaxconn`, we'll consider moving back to the unix domain socket, but for now I'd like to use TCP socket, which doesn't have such a low limit.

bshaffer:
aha, got it, ty! LGTM

tmatsuo:
Thanks!

cedricziel:
Did you check if the issues persist when using the retry and timeout you added with the socket?

tmatsuo:
@cedricziel I thought I checked, but apparently I haven't. Thanks for the suggestion!

Actually using the unix socket with the retry gave me ~3 times better performance than using tcp socket with no errors.

I'll go back to unix socket soon.

Thanks again!!

cedricziel:
Good to hear :)

tmatsuo:
@cedricziel I should have tested more before replying you.

Nah, it still return 502 when receiving lots of concurrent requests. Obviously unix socket is faster, but I think it's important to prevent errors as much as possible. I will explore how to change the kernel parameter, and will move back to unix socket once I know how.

Anyways, thanks for your opinion!

PR #74, opened 2016-03-17
Removed libxml_disable_entity_loader from the blacklist.

### LINKED ISSUE #73: Symfony requires function "libxml_disable_entity_loader"
The function [libxml_disable_entity_loader](http://php.net/manual/en/function.libxml-disable-entity-loader.php) is required in Symfony's XMLFileLoader, and is blacklisted in [our configuration](https://github.com/GoogleCloudPlatform/php-docker/blob/master/php-nginx/php.ini#L938).

Symfony is using the function to make its xml file loading more safe. Wouldn't it be better to call the function (`libxml_disable_entity_loader(true);`) than disable it? Can suhosin be used to disable the entity loader at the php.ini level, thus rendering the function useless?

tmatsuo:
Fixes #73

### DIFF
--- a/php-nginx/php.ini
+++ b/php-nginx/php.ini
@@ -935,7 +935,7 @@ extension=memcached.so
 
 ; Configure the suhosin defaults that we want.
 suhosin.executor.disable_emodifier = On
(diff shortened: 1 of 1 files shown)


bshaffer:
LGTM, we can use [auto_prepend_file](http://docs.php.net/manual/en/ini.core.php#ini.auto-prepend-file) in `php.ini` to add a global bootstrap for calling this function.

bshaffer:
We should also remove [`parse_str`](http://php.net/manual/en/function.parse-str.php) from this list while we are at it. I do not understand why this would be blacklisted, it is a *very* common function, and not at all unsafe. 

I believe it's because of the functionality when `parse_str` is called without a second argument:

```php
$var1 = 1;
parse_str('var1=one&var2=two');
// $var1 is now 'one'
```

But this is extremely restrictive.

tmatsuo:
Restarted the build and it should pull #77 and the test should pass

tmatsuo:
@bshaffer 

Well, considering that `parse_str` can override the local scope variable as well as global variables, I hesitate to remove it from the blacklist.

tmatsuo:
Merging. Please open a new issue if you still think we should remove `parse_str` from the blacklist.

PR #76, opened 2016-03-17
Changed the file owner to root.www-data.

### LINKED ISSUE #75: Application files should be owned by root
Now we have the following in the `Dockerfile`:
```
ONBUILD RUN chown -R www-data.www-data $APP_DIR
```
It should be root.www-data so that php can not override any files.

tmatsuo:
Fixes #75

### DIFF
--- a/php-nginx/Dockerfile
+++ b/php-nginx/Dockerfile
@@ -124,7 +124,7 @@ RUN cd /tmp && ${PHP_DIR}/bin/php \
 
 ONBUILD COPY . $APP_DIR
 ONBUILD RUN chmod -R 550 $APP_DIR
-ONBUILD RUN chown -R www-data.www-data $APP_DIR
+ONBUILD RUN chown -R root.www-data $APP_DIR
 
 WORKDIR $APP_DIR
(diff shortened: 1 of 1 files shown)


bshaffer:
Looks good, although symfony will need to write the cache, so this change will break [this](https://github.com/GoogleCloudPlatform/php-docs-samples/blob/master/managed_vms/symfony/Dockerfile#L9)

tmatsuo:
Yes, probably you need g+w?

cedricziel:
In a derived image I use with symfony I needed to completely reset the file modes and ownerships to `www-data` + the umask (to 002) of the fpm process to have have a stable experience.

tmatsuo:
@cedricziel alright, can you share your Dockerfile?

tmatsuo:
Restarted the travis build

tmatsuo:
Yeah it may break some apps. Do you mind if I submit this and push the latest image (it will happen automatically)?

tmatsuo:
Ok, I would go ahead and merge this.

PR #77, opened 2016-03-17
Allowing users to override php-fpm.conf.

### LINKED ISSUE #71: The builds on travis are flaky
while builds on jenkins are stable

tmatsuo:
Also stopped using keepalive for fcgi in nginx.conf for test stability on travis, added debug output for local docker tests.

Fixes #71 
Related to #70

### DIFF
--- a/php-nginx/Dockerfile
+++ b/php-nginx/Dockerfile
@@ -103,6 +103,7 @@ COPY php.ini $PHP56_DIR/lib/php.ini
 COPY php.ini $PHP7_DIR/lib/php.ini
 COPY php-fpm.conf $PHP56_DIR/etc/php-fpm.conf
 COPY php-fpm.conf $PHP7_DIR/etc/php-fpm.conf
+RUN touch $PHP56_DIR/etc/php-fpm-user.conf $PHP7_DIR/etc/php-fpm-user.conf
 COPY supervisord.conf /etc/supervisor/supervisord.conf
 COPY logrotate.app_engine /etc/logrotate.d/app_engine
 
--- a/php-nginx/entrypoint.sh
+++ b/php-nginx/entrypoint.sh
@@ -28,6 +28,9 @@ APP_NGINX_ADDITIONAL_CONF=${APP_DIR}/nginx-app.conf
 # replace our default main configuration file with this file.
 APP_NGINX_CONF=${APP_DIR}/nginx.conf
 
+# User provided php-fpm.conf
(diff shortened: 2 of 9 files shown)


tmatsuo:
I think I found possible cause. The test is failing with the following error on the server:

```
  string(255) "2016/03/17 18:08:44 [error] 18#0: *5 upstream sent unexpected FastCGI record: 3 while reading response header from upstream, client: 172.17.42.1, server: , request: "GET /phpinfo.php HTTP/1.1", upstream: "fastcgi://127.0.0.1:9000", host: "localhost:65080""
```

This might be a known issue with `keepalive` setting in nginx.conf.
https://github.com/perusio/drupal-with-nginx/issues/160
https://groups.drupal.org/node/283558#comment-1016818

I will remove the `keepalive` setting and see how it goes.

tmatsuo:
Ok, the build started to pass. I squashed the commits.
@bshaffer can you take a look?

php-nginx/Dockerfile
```diff
@@ -103,6 +103,7 @@ COPY php.ini $PHP56_DIR/lib/php.ini
 COPY php.ini $PHP7_DIR/lib/php.ini
 COPY php-fpm.conf $PHP56_DIR/etc/php-fpm.conf
 COPY php-fpm.conf $PHP7_DIR/etc/php-fpm.conf
+RUN touch $PHP56_DIR/etc/php-fpm-user.conf $PHP7_DIR/etc/php-fpm-user.conf
```
bshaffer: Why do we `touch` these files after copying them?

bshaffer:
I'd like to see a test, otherwise LGTM

php-nginx/Dockerfile
cedricziel: Those are the user confs, not the main ones from above. Touching them shall prevent warnings I guess.

bshaffer: cool, thanks :+1:

tmatsuo: Yes, just making sure the included files are there.

tmatsuo:
Ok, added simple tests.

bshaffer:
lookin' good

tmatsuo:
Thanks @bshaffer 

We need to count master process + child processes. So if you limit the max_children = 1, there should be 2 processes (1 master and 1 child) in the docker image.

This time test should pass!

tmatsuo:
Algirht, merging!

PR #85, opened 2016-03-31
Patches parse_str to make the 2nd parameter mandatory.

### LINKED ISSUE #79: remove parse_str from blacklist
We should remove [`parse_str`](http://php.net/manual/en/function.parse-str.php) from the suhosin blacklist.

As @tmatsuo mentioned, if used in a certain way, it's possible for this function to override global variables. However, we need to keep in mind:

1. The function can be used safely when a second argument is supplied
2. The function is very common, as it's the only way (other than rolling your own) to parse a query string.
3. The dangers of using this function require the user to be doing a few other things incorrectly to do damage, namely not providing a 2nd argument to `parse_str` and relying on global scope throughout the app. 

It would be best to remove this function from the blacklist. The best solution would be to disable the function's ability to be used without a second argument, but this is out of scope.

tmatsuo:
Fixes #79

### DIFF
--- a/php-nginx/Dockerfile
+++ b/php-nginx/Dockerfile
@@ -78,11 +78,15 @@ RUN /bin/bash /build-scripts/build_nginx.sh
 # Build PHP 5.6
 ENV PHP56_VERSION=5.6.19
 COPY build-scripts/build_php56.sh /build-scripts/build_php56.sh
+COPY build-scripts/php5-parse_str_harden.patch \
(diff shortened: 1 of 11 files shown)


tmatsuo:
@raggi I will appreciate it if you can take a look.

For background, see discussion on #79

raggi:
LGTM

tmatsuo:
Thanks!

PR #91, opened 2016-04-29
Upgrade PHP to 5.6.21 & 7.0.6

cedricziel:
We should think about a way to store away the precompiled binaries in a trusted way somehow and only compile them when needed.


cedricziel:
Once again the PGP Servers are down or deny traffic from travis or whatever. -.-

Feel free to restart the build in a while.

tmatsuo:
Thanks @cedricziel !

PR #92, opened 2016-05-21
Combine some Docker instructions for making the image smaller.

tmatsuo:
Also minor version updates.

The image size was ~830MB, but with this change, now it's ~650MB.

### DIFF
--- a/php-nginx/build-scripts/build_php56.sh
+++ b/php-nginx/build-scripts/build_php56.sh
@@ -47,7 +47,7 @@ rm mailparse.tar.gz
 
 mkdir -p ${PHP_SRC}/ext/apcu
 # The newest 5.1.2 doesn't build with PHP 5.6.
-curl -SL "https://pecl.php.net/get/apcu-4.0.10.tgz" -o apcu.tar.gz
+curl -SL "https://pecl.php.net/get/apcu-4.0.11.tgz" -o apcu.tar.gz
 tar -zxf apcu.tar.gz -C ${PHP_SRC}/ext/apcu --strip-components=1
 rm apcu.tar.gz
 
--- a/php-nginx/Dockerfile
+++ b/php-nginx/Dockerfile
@@ -53,43 +53,21 @@ ENV NGINX_DIR=/usr/local/nginx \
     UPLOAD_DIR=/upload \
     SESSION_SAVE_PATH=/tmp/sessions \
     OPENSSL_VERSION=1.0.1p \
+    NGINX_VERSION=1.10.0 \
+    PHP56_VERSION=5.6.21 \
+    PHP70_VERSION=7.0.6 \
     PATH=/usr/local/php/bin:$PATH
 
 # BUILD PHP, nginx and other dependencies.
 COPY openssl-version-script.patch /tmp/openssl-version-script.patch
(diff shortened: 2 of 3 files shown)


cedricziel:
I'm generally fine with this. What's the virtual size of the base image?

php-nginx/build-scripts/build_php56.sh
```diff
@@ -47,7 +47,7 @@ rm mailparse.tar.gz
 
 mkdir -p ${PHP_SRC}/ext/apcu
 # The newest 5.1.2 doesn't build with PHP 5.6.
-curl -SL "https://pecl.php.net/get/apcu-4.0.10.tgz" -o apcu.tar.gz
+curl -SL "https://pecl.php.net/get/apcu-4.0.11.tgz" -o apcu.tar.gz
```
JustinBeckwith: total nit - but shouldn't we be doing version updates in a different PR?

tmatsuo: I separated the changes

tmatsuo:
ugh, github doesn't pick up my changes, I will delete it and open a new one

tmatsuo:
Sorry guys the new one is here:
https://github.com/GoogleCloudPlatform/php-docker/pull/94

tmatsuo:
@cedricziel Do you mean "gcr.io/google_appengine/base"? Then it's 246.1 MB

cedricziel:
@tmatsuo Thank you for getting back to me :) 

Yes - that's a huge difference; did you spend some thoughts on the initial idea of maybe being able to split out the static compiled PHP Versions? I could imagine being able to push those as artifacts to GCS. On build, the user could decide which version of the interpreter and which version and / or flavor of http-daemon (nginx|apache|hhvm) should be pulled in.

That could make for a lot smaller images and therefore faster deployments. Not to mention the flexibility.

raggi:
There have been proposals internally to start a new convention of providing versioned image names, for example:
`gcr.io/google_appengine/php-nginx-5.6`
We haven't gotten to do this yet, or even to ratify such a proposal, but it's been discussed quite a few times. Personally I think we should do this.

Regarding pushing artifacts to GCS and adding them on build. I'm not sure this would have exactly the effect you desire, in that, this would result in your builds generating much larger personal artifacts. You would be generating a user owned layer in your image that contains the result of the installation of that artifact.

Now, again, I have proposed and will keep pushing for us actually building the various non-standard packages as .deb files, and separating the builds of those out from the runtime Dockerfiles. This should improve many things, not least of which would be our end to end test performance for the runtime docker images and more repeatable builds at specific layers. Personally I don't think it would be ideal to have users have to install those debs themselves (unless they're using runtime:custom). Instead I would prefer to handle major versions and different options (such as nginx for php) in terms of the docker image name. Taking this approach minimizes the time taken to fetch the base runtime image during a deploy, and reduces the size of the user image layers when they're pushed after a build.

raggi:
@JustinBeckwith Take a look at the above comment, this plays again into many of our discussions :)

cedricziel:
Sure that's another way to put it and having different names for different versions isn't a bad idea either. I just wanted to make sure it's still on the page.

PR #94, opened 2016-05-23
Combine some Docker instructions for making the image smaller

tmatsuo:


php-nginx/Dockerfile
```diff
@@ -53,43 +53,21 @@ ENV NGINX_DIR=/usr/local/nginx \
...
-
-# Build PHP 7.0
-ENV PHP70_VERSION=7.0.6
-COPY build-scripts/build_php70.sh /build-scripts/build_php70.sh
-COPY build-scripts/php7-parse_str_harden.patch \
-    /build-scripts/php7-parse_str_harden.patch
-RUN /bin/bash /build-scripts/build_php70.sh
-
-RUN /bin/bash /build-scripts/apt_build_deps.sh uninstall
+RUN /bin/bash /build-scripts/apt_build_deps.sh install && \
```
raggi: We probably want to discuss this build strategy in general. It results in very little reuse / layering, which might be what we want, but there are pro's and cons. Other runtimes are not taking this approach.

raggi:
LGTM

bshaffer:
Late to the party, but LGTM

PR #104, opened 2016-06-02
Allow long requests by default.

tmatsuo:
Also fixes a bug about wrong nginx.conf path.


tmatsuo:
The test for long running request was failing because on travis, guzzle uses fopen and the default_socket_timeout configuration (default 60 seconds) kicks in. I added a configuration change in the test code and let's see if it passes.

bshaffer:
LGTM!

tmatsuo:
I'll let you know once it's uploaded to prod

tmatsuo:
It's uploaded

bshaffer:
woohoo! I'll check it out now.

bshaffer:
it worked!

PR #105, opened 2016-06-02
Embedded public keys for verifying the tarballs.

### LINKED ISSUE #97: Remove PGP key import script
It's not being used, and should be removed.

tmatsuo:
fixes #97


php-nginx/Dockerfile
```diff
@@ -57,15 +57,18 @@ ENV NGINX_DIR=/usr/local/nginx \
...
 COPY build-scripts /build-scripts
 
 RUN /bin/bash /build-scripts/apt_build_deps.sh install && \
-    /bin/bash /build-scripts/import_pgp_keys.sh && \
     /bin/bash /build-scripts/build_nginx.sh && \
     /bin/bash /build-scripts/build_php56.sh && \
     /bin/bash /build-scripts/build_php70.sh && \
-    /bin/bash /build-scripts/apt_build_deps.sh uninstall
+    /bin/bash /build-scripts/apt_build_deps.sh uninstall && \
+    rm -rf /build-scripts /gpgkeys
```
raggi: The content of the scripts will remain present anyway, so I'm not sure there's a good reason to remove them here.

raggi:
Other than the nit mentioned, this looks good.

php-nginx/Dockerfile
tmatsuo: Good point, it doesn't give us anythiing. I removed it.

tmatsuo:
I'll merge this once the test passes.

PR #106, opened 2016-06-03
Use /opt instead of /usr/local.

### LINKED ISSUE #96: Move the manual builds into /opt instead of /usr/local
It's odd to build into /usr/local/<package>, we should use /opt/<package> if we're not integrating into /usr/local.

tmatsuo:
fixes #96


raggi:
LGTM

tmatsuo:
@cedricziel FYI, php and nginx will be moved to /opt

PR #108, opened 2016-06-04
Env vars to change the custom config file name.

### LINKED ISSUE #102: Difficulties with more than one service per app
It is difficult to use this runtime to deploy more than one service per app to AppEngine, due to the requirement of each service to share `nginx-app.conf` `php.ini` and `supervisord.conf` files.

We could add a `runtime_config` variable `conf_dir` that can be used to change the configuration dir from the default (`/app`) to something else (like `/app/config/my_service`)

tmatsuo:
fixes #102

### DIFF
--- a/php-nginx/README.md
+++ b/php-nginx/README.md
@@ -246,6 +246,8 @@ There are environment variables for adding config files.
   Additional php config file.
 - SUPERVISORD_CONF_ADDITION (defaults to additional-supervisord.conf)
   Additional supervisord config file.
+- SUPERVISORD_CONF_OVERRIDE (defaults to supervisord.conf)
+  The main supervisord config file.
 
 If those files are present, the runtime will copy those files to
 appropriate place.
--- a/php-nginx/README.md
+++ b/php-nginx/README.md
@@ -236,15 +236,15 @@ various config files" below).
 
 There are environment variables for adding config files.
 
-- APP_NGINX_ADDITIONAL_CONF (defaults to nginx-app.conf)
+- NGINX_CONF_INCLUDE (defaults to nginx-app.conf)
   This file will be inserted to the main nginx configuration.
-- APP_NGINX_CONF (defaults to nginx.conf)
+- NGINX_CONF_OVERRIDE (defaults to nginx.conf)
   This file will entirely override the main nginx configuration.
-- APP_PHP_FPM_CONF (defaults to php-fpm.conf)
+- PHP_FPM_CONF_OVERRIDE (defaults to php-fpm.conf)
   Additional config file for php-fpm.
-- APP_PHP_INI (defaults to php.ini)
+- PHP_INI_OVERRIDE (defaults to php.ini)
   Additional php config file.
-- APP_SUPERVISORD_CONF (defaults to supervisord.conf)
+- SUPERVISORD_CONF_ADDITION (defaults to additional-supervisord.conf)
   Additional supervisord config file.
 
 If those files are present, the runtime will copy those files to
--- a/php-nginx/README.md
+++ b/php-nginx/README.md
@@ -229,7 +229,26 @@ These extensions are only available with PHP 7:
 ## Add something to php.ini
 
 You can just have php.ini file in your project directory. This file is
-added to the main configuration file.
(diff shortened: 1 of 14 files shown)


bshaffer:
This is amazing and brings a tear to my eye. Nice work!

tmatsuo:
Looking at it now, I kind of feel the env names are not consistent and verbose.

How about changing then like:

`APP_NGINX_ADDITIONAL_CONF` -> `NGINX_CONF_INCLUDE`
`APP_NGINX_CONF` -> `NGINX_CONF_OVERRIDE`
`APP_PHP_FPM_CONF` -> `PHP_FPM_CONF_OVERRIDE`
`APP_PHP_INI` -> `PHP_INI_OVERRIDE`
`APP_SUPERVISORD_CONF` -> `SUPERVISORD_CONF_ADDITION`

?

tmatsuo:
That way, the envvar represents what they are.

For example, in the future it's easy to add `SUPERVISORD_CONF_OVERRIDE` which is a complete replacement of the main supervisord config.

bshaffer:
Ahh, I think I like this. But what's the difference between `ADDITION` and `INCLUDE` here?

On another note, I actually need the supervisord override and not the addition.

On another another note, the `PATH` declaration does not seem to be picking up `${PHP_DIR}/bin`, and I'm not sure why, as it's clearly defined [here](https://github.com/GoogleCloudPlatform/php-docker/blob/master/php-nginx/Dockerfile#L58)

tmatsuo:
> Ahh, I think I like this. But what's the difference between ADDITION and INCLUDE here?
INCLUDE is included into the main config file
ADDITION is a file that will be read additionally to the main config file

> On another note, I actually need the supervisord override and not the addition.
Little bit questionable to support that case with this image (user can easily hit their feet).

> On another another note, the PATH declaration does not seem to be picking up ${PHP_DIR}/bin, and I'm not sure why, as it's clearly defined here

Probably it's because it's reset by bash (/etc/profile)?

bshaffer:
I think it would be best to just keep it simple and have the supervisord be an override. Unless there's another way to not run `php-fpm`.

tmatsuo:
You can configure php-fpm to run only small amount of instances, or zero.

bshaffer:
If you're referring to `process.max`, a value of zero denotes an unlimited amount.

tmatsuo:
pm = ondemand

then there will be no processes until they are needed

bshaffer:
I doubt that will stop nginx from listening on port `8080`

tmatsuo:
In most cases, use nginx for health check

bshaffer:
OK, LGTM! I think this will work.

bshaffer:
Do we want the copying/moving of these files to happen in `entrypoint.sh`? It seems like this would be better off in the docker image. I assume this has to be done in `entrypoint.sh` because of the environment variables.

tmatsuo:
AFAIK, this is the only way right now, but I'm open to other optoins.

tmatsuo:
We chatted offline. Maybe we can introduce `SUPERVISORD_CONF_OVERRIDE`.

tmatsuo:
FYI, this change is released on gcr.io/php-mvm-a/php-nginx as well as gcr.io/google_appengine/php with a release tag `r20160606`.

cedricziel:
This change breaks custom module requirements for composer installations which are run way before entrypoint is ever executed.

I suggest to add the app dir back to conf-scan. Otherwise we raise an additional hurdle requiring the user to COPY the custom ini to the scan dir through the dockerfile (which would make this change superfluous.

cedricziel:
I figured I'd illustrate this a bit:

* composer.json has a platform requirement on ext-gd
* image is built and there is **NO** way to add the gd module to any of the inis, because the first composer execution is an ONBUILD one where only the inis from the initial image build are present
* build fails since composer install fails

cedricziel:
Correction: There is no way to ever add modules at the needed point in time. I proposed #109 to re-add the ini scan.

tmatsuo:
Sorry for the breakage. Merged and building the image for upload.

cedricziel:
No worries - thanks for the quick merges!

PR #109, opened 2016-06-07
Re-Add APP_DIR to ini scan folders

cedricziel:
GoogleCloudPlatform/php-docker#108 removed the possibility
to add PHP modules on build, which fails composer when there's
platform requirements such as `ext-gd`.

This patch re-adds the additonal scan folder.

PR #110, opened 2016-06-07
Add php-cs-fixer through project level composer dependencies

cedricziel:
According to FriendsOfPHP/PHP-CS-Fixer#1966, the availability of the
2.0-dev version as phar from dist was an accident.

This change adds the 2.0-alpha version through the project level dependencies.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,5 +1,6 @@
 {
     "require": {
+        "friendsofphp/php-cs-fixer": "v2.0.0-alpha"
     },
     "require-dev": {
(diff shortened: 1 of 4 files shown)


cedricziel:
This should fix the build errors from #109 (please have a look at the linked ticket from php-cs-fixer there).

tmatsuo:
LGTM

tmatsuo:
On the second thought, we should use the stable phar. I can merge this though.

cedricziel:
Thx! :)

PR #115, opened 2016-06-07
Use php-cs-fixer 1.11.3

tmatsuo:


tmatsuo:
I needed to use php-cs-fixer 1.11.3 due to the old version of PHP on our jenkins server.
Also needed to use --prefer-lowest to create the composer.lock file for the same reason.

bshaffer:
Why can't we just use composer to install the required version? I don't see the advantage with this change.

bshaffer:
i.e just use:
```php
"friendsofphp/php-cs-fixer": "1.11.3",
```

from [this tag](https://github.com/FriendsOfPHP/PHP-CS-Fixer/tree/v1.11.3)

tmatsuo:
Yeah, that works too

tmatsuo:
@bshaffer PTAL

bshaffer:
LGTM

PR #118, opened 2016-06-14
fixes PHP_DIR in PATH

bshaffer:
`PATH=${PHP_DIR}/bin:$PATH` needs to be on a separate line, as it uses env variables declared in the same line.

### DIFF
--- a/php-nginx/Dockerfile
+++ b/php-nginx/Dockerfile
@@ -55,7 +55,7 @@ ENV NGINX_DIR=/opt/nginx \
     NGINX_VERSION=1.10.1 \
     PHP56_VERSION=5.6.22 \
     PHP70_VERSION=7.0.7 \
-    PATH=${PHP_DIR}/bin:$PATH
+    PATH=/opt/php/bin:$PATH
 
 # gpgkeys for verifying the tarballs
 COPY gpgkeys /gpgkeys
--- a/tests/VersionTest.php
+++ b/tests/VersionTest.php
@@ -59,4 +59,11 @@ class VersionTest extends \PHPUnit_Framework_TestCase
         }
         $this->assertTrue($checked, 'Failed to check the version is 5.6.x');
     }
+    public function testPhpInPath()
+    {
+        exec('docker run php56_70 env | grep "^PATH="', $output);
+        $grep = array_pop($output);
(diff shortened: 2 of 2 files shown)


php-nginx/Dockerfile
```diff
@@ -54,8 +54,9 @@ ENV NGINX_DIR=/opt/nginx \
     SESSION_SAVE_PATH=/tmp/sessions \
     NGINX_VERSION=1.10.1 \
     PHP56_VERSION=5.6.22 \
-    PHP70_VERSION=7.0.7 \
-    PATH=${PHP_DIR}/bin:$PATH
+    PHP70_VERSION=7.0.7
+
+ENV PATH=${PHP_DIR}/bin:$PATH
```
tmatsuo: I would prefer hard coding the path. A new instruction in the Dockerfile introduce another docker layer which will increase the image size a bit.

tmatsuo:
Oh good catch :)

php-nginx/Dockerfile
bshaffer: should we just hardcode it then? 

```
ENV NGINX_DIR=/opt/nginx \
    PHP_DIR=/opt/php \
    // ....
    PATH=/opt/php/bin:$PATH
```

tmatsuo: Yeah, that's my preference.

tmatsuo:
https://github.com/GoogleCloudPlatform/php-docker/pull/119 will fix the build

tmatsuo:
Unfortunately the build will fail unless you merge #119 first

tmatsuo:
LGTM though

tmatsuo:
Restarted the build, it should pass now

tmatsuo:
FYI, the test you added is failing.

tmatsuo:
Does it work if you do:
```
exec('docker run php56_70 env|grep ^PATH');
```
instead of
```
exec('docker run php56_70 echo $PATH');
```
?

You can also test it on local machine easily.

tmatsuo:
Alright, I'll let you know once the image is in production

tmatsuo:
The image has been pushed to `gcr.io/google_appengine/php:latest` as well as `gcr.io/php-mvm-a/php-nginx:latest`

PR #119, opened 2016-06-14
Pin redis to 2.2.8 for PHP 5.6.

tmatsuo:


tmatsuo:
It should fix the build failure

tmatsuo:
@bshaffer The new redic pecl package only supports PHP >= 7.0, so I needed to pin redis version

bshaffer:
LGTM!

PR #120, opened 2016-06-15
Removed glcoud config set app/cloud_build

tmatsuo:
The new cloud SDK removed this config and it's erroring


JustinBeckwith:
LGTM

tmatsuo:
Merging this without running test, because there are 2 issues affecting the build

PR #121, opened 2016-06-18
Added an envvar WHITELIST_FUNCTIONS.

tmatsuo:
To create appropriate ini file at runtime.

### DIFF
--- a/php-nginx/whitelist_functions.php
+++ b/php-nginx/whitelist_functions.php
@@ -1,11 +1,5 @@
 <?php
 
-function filter_blacklist($val)
-{
-    global $whitelist;
-    return !in_array($val, $whitelist);
-}
-
 $phpDir = getenv("PHP_DIR");
 
 if ($phpDir === false) {
@@ -32,7 +26,7 @@ if ($suhosinBlacklist !== false) {
         "suhosin.executor.func.blacklist=\"%s\"\n",
         implode(
             ',',
-            array_filter($suhosinBlacklist, 'filter_blacklist')
+            array_diff($suhosinBlacklist, $whitelist)
         )
     );
 }
@@ -44,7 +38,7 @@ if ($disable_functions !== false) {
         "disable_functions = \"%s\"\n",
         implode(
             ',',
-            array_filter($disable_functions, 'filter_blacklist')
+            array_diff($disable_functions, $whitelist)
         )
     );
 }
--- a/php-nginx/whitelist_functions.php
+++ b/php-nginx/whitelist_functions.php
@@ -1,0 +1,57 @@
+<?php
+
+function filter_blacklist($val)
+{
+    global $whitelist;
+    return !in_array($val, $whitelist);
+}
+
+$phpDir = getenv("PHP_DIR");
+
+if ($phpDir === false) {
+    exit(0);
+}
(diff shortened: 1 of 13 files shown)


php-nginx/whitelist_functions.php
```diff
@@ -0,0 +1,57 @@
...
+}
+
+$suhosinBlacklist = ini_get('suhosin.executor.func.blacklist');
+$disable_functions = ini_get('disable_functions');
+
+$whitelist_ini = '';
+
+if ($suhosinBlacklist !== false) {
+    $suhosinBlacklist = array_filter(
+        array_map('trim', explode(',', $suhosinBlacklist)));
```
bshaffer: just use `array_diff` here instead of defining a custom function, global variables, and using array_map, i.e:

```php
$whitelist_ini .= sprintf(
         "suhosin.executor.func.blacklist=\"%s\"\n",
         implode(
             ',',
             array_diff($suhosinBlacklist, $whitelist)
         )
     );
```

bshaffer:
I like this. I don't think we should require `gcloud` for this.

tmatsuo:
FYI, the travis hook is disables for a time being. Running a build on our jenkins server now.

tmatsuo:
@bshaffer 

Sorry for the ambiguity. I said gcloud, referring to gcloud SDK. Anyways, thinking it now, we run composer with the following:

```
        -d disable_functions= \
```

So my worry is not a real issue. Also since we have nice set of tests, it should be fine in terms of. I removed [Don't merge] from the title. Also, the tests are passing on our jenkins.

@bshaffer PTAL

php-nginx/whitelist_functions.php
```diff
@@ -0,0 +1,51 @@
...
+} else {
+    $whitelist = array_filter(array_map('trim', explode(',', $whitelist)));
+}
+
+$suhosinBlacklist = ini_get('suhosin.executor.func.blacklist');
+$disable_functions = ini_get('disable_functions');
+
+$whitelist_ini = '';
+
+if ($suhosinBlacklist !== false) {
```
bshaffer: So `WHITELIST_FUNCTIONS` trumps `suhosin.executor.func.blacklist` and `disable_functions`. I think this makes sense.

testapps/php56_custom/web/phpinfo.php
```diff
@@ -15,4 +15,9 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+
+// exec is disabled by default, but it should run
+if (exec('echo hello') !== 'hello') {
```
bshaffer: nit: Why not make a new `exec.php` for this test?

tmatsuo: Done

tmatsuo: Done

tmatsuo:
@bshaffer PTAL

tmatsuo:
The tests are passing on our jenkins :)

bshaffer:
LGTM!

tmatsuo:
Thanks!

tmatsuo:
JFYI, It's pushed to production!

PR #122, opened 2016-06-25
Update PHP versions to 5.6.23  and 7.0.8

cedricziel:


tmatsuo:
Sorry for the late response, merging.

tmatsuo:
The change is now published to production

cedricziel:
Thx! No need for apologizing :)

PR #123, opened 2016-06-28
adds ignore-platform-reqs to composer install

bshaffer:


bshaffer:
because `php.ini` settings may be loaded in `/entrypoint.sh` via the custom `PHP_INI_OVERRIDE` environment variable, we cannot throw platform requirement errors in `./composer.sh`.

cedricziel:
Please dont.

The platform requirements enabled installation acts as a saveguard, and when it's gone, there's none. As I've mentioned several times before, I'm not satisfied with the current state as well - but please, let's not make it worse.

tmatsuo:
Can we close this?

PR #124, opened 2016-07-08
Adding circleci configuration.

tmatsuo:
Migrating to circleci for PR hook.

### DIFF
--- a/scripts/dump_credentials.php
+++ b/scripts/dump_credentials.php
@@ -21,6 +21,9 @@
  *
  * To setup Travis to run on your fork, read TRAVIS.md.
  */
+if (getenv('GOOGLE_CREDENTIALS_BASE64') === false) {
+    exit(0);
+}
 file_put_contents(
     getenv('PHP_DOCKER_GOOGLE_CREDENTIALS'),
     base64_decode(getenv('GOOGLE_CREDENTIALS_BASE64'))
--- a/scripts/install_test_dependencies.sh
+++ b/scripts/install_test_dependencies.sh
@@ -15,9 +15,12 @@
 
 set -ex
 
-if [ "${CIRCLECI}" != "true" ]; then
+if [ "${INSTALL_PHP5}" == "true" ]; then
     sudo apt-get update
     sudo apt-get install -y php5-cli
+fi
+
+if [ "${INSTALL_GCLOUD}" == "true" ]; then
     # Install gcloud
     if [ ! -d ${HOME}/gcloud/google-cloud-sdk ]; then
         mkdir -p ${HOME}/gcloud &&
@@ -35,7 +38,7 @@ if [ -z "${CLOUDSDK_ACTIVE_CONFIG_NAME}" ]; then
 fi
 
 # Install composer and defined dependencies
-curl -sS https://getcomposer.org/installer | sudo php -- --install-dir=/usr/local/bin --filename=composer
+which composer || curl -sS https://getcomposer.org/installer | sudo php -- --install-dir=/usr/local/bin --filename=composer
 composer install --ignore-platform-reqs
 
 # gcloud configurations
--- a/scripts/install_test_dependencies.sh
+++ b/scripts/install_test_dependencies.sh
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
 # Copyright 2015 Google Inc.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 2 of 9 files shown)


tmatsuo:
@waprin FYI
Also I will appreciate it if you LGTM too since both brent and jeff is out now

scripts/install_test_dependencies.sh
```diff
@@ -15,8 +15,19 @@
 
 set -ex
 
-sudo apt-get update
-sudo apt-get install -y php5-cli
+if [ "${CIRCLECI}" != "true" ]; then
+    sudo apt-get update
+    sudo apt-get install -y php5-cli
+    # Install gcloud
```
waprin: believe gcloud comes pre-installed on circle

waprin: oh now I see you're aware and want your own gcloud...but why?

scripts/install_test_dependencies.sh
```diff
@@ -15,8 +15,19 @@
...
+    if [ ! -d ${HOME}/gcloud/google-cloud-sdk ]; then
+        mkdir -p ${HOME}/gcloud &&
+        wget https://dl.google.com/dl/cloudsdk/release/google-cloud-sdk.tar.gz --directory-prefix=${HOME}/gcloud &&
+        cd "${HOME}/gcloud" &&
+            tar xzf google-cloud-sdk.tar.gz &&
+            ./google-cloud-sdk/install.sh --usage-reporting false --path-update false --command-completion false &&
+            cd "${TRAVIS_BUILD_DIR}";
```
waprin: Don't think TRAVIS_BUILD_DIR is right?

scripts/install_test_dependencies.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
```
waprin: think "#!/usr/bin/env bash " is preferred if we're going to fix it.

scripts/install_test_dependencies.sh
tmatsuo: I wanted to use the latest gcloud sdk, not the preinstalled one. I'm fine with upgrading on demand like I do now.

I prefer using our own installation because potentially I can test whatever version including prerelease, specific version, etc.

scripts/install_test_dependencies.sh
tmatsuo: Yeah I set it on circle.yml, but it's weird. I changed it to TEST_BUILD_DIR.

scripts/install_test_dependencies.sh
tmatsuo: Done

tmatsuo:
@waprin Thanks! PTAL

waprin:
lgtm

tmatsuo:
Thanks, will merge once the test passes on circle

tmatsuo:
Unfortunately `docker exec` is not supported on circleci, so our tests are failing

tmatsuo:
Workaround:
https://circleci.com/docs/docker/#docker-exec

tmatsuo:
Or, stop using `docker exec` and create a web handler that reports the number of processes of nginx and php-fpm. I prefer this approach.

tmatsuo:
I think we need a reliable way to wait for the nginx to start, instead of sleeping arbitrary number of seconds.

tmatsuo:
Yes, the tests are now passing on circleci!

scripts/dump_credentials.php
```diff
@@ -21,6 +21,9 @@
  *
  * To setup Travis to run on your fork, read TRAVIS.md.
  */
+if (getenv('GOOGLE_CREDENTIALS_BASE64') === false) {
+    exit(0);
```
bshaffer: I am glad to see this step is optional

scripts/install_test_dependencies.sh
bshaffer: Looks like this code path only execute if we aren't on CircleCI. What is this for?

tmatsuo: On CircleCI, we have PHP 5.6.22 by default and it works pretty well for us.

scripts/install_test_dependencies.sh
```diff
@@ -15,8 +15,19 @@
 
 set -ex
 
-sudo apt-get update
-sudo apt-get install -y php5-cli
+if [ "${CIRCLECI}" != "true" ]; then
+    sudo apt-get update
+    sudo apt-get install -y php5-cli
```
bshaffer: I feel like these shouldn't be here. It ruins this script for OSX and some versions of linux. Can we make it more generic?

tmatsuo: This script is only for CI systems. You don't have to run this script locally. Locally you can install them in whatever way you like.

Are you comfortable if it runs only on Jenkins?

tmatsuo: @bshaffer Ok, I made it harmless. PTAL

tmatsuo:
Rebased to the current master

tmatsuo:
@bshaffer The test is still passing. Anything left?

bshaffer:
This is awesome, LGTM

tmatsuo:
Thanks!

PR #127, opened 2016-07-12
Use `gcloud app` instead of `gcloud preview app`

tmatsuo:
We should use `gcloud app` instead of `gcloud preview app` which will go away soon.

### DIFF
--- a/php-nginx/README.md
+++ b/php-nginx/README.md
@@ -264,7 +264,7 @@ Nginx and php-fpm are communicating via the TCP port 9000.
 You can deploy the app by:
 
 ```sh
-$ gcloud preview app deploy app.yaml
+$ gcloud app deploy
 ```
 
 ## Run it locally with docker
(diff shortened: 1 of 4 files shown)


php-nginx/README.md
```diff
@@ -280,7 +280,7 @@ Nginx and php-fpm are communicating via the TCP port 9000.
 You can deploy the app by:
 
 ```sh
-$ gcloud preview app deploy app.yaml
+$ gcloud app deploy app.yaml
```
bshaffer: We should get rid of `app.yaml` here also, since it's no longer necessary

tmatsuo: Done

bshaffer:
LGTM once circleci passes!

tmatsuo:
Yeah #124 needs to be merged for this PR to pass on circleci

tmatsuo:
Rebased and made some tweaks on READMEs.

tmatsuo:
We needed to skip the e2e test when secrets are no available, and it should pass this time!

tmatsuo:
Looking good, merging

PR #128, opened 2016-07-18
Do not set HTTP_PROXY environment variable.

tmatsuo:


tmatsuo:
I will add the test in the later CL

broady:
LGTM

```
Nginx/FastCGI
Use this to block the header from being passed on to PHP-FPM, PHP-PM etc.

fastcgi_param HTTP_PROXY "";
In FastCGI configurations, PHP is vulnerable (but many other languages that use Nginx FastCGI are not).
```

bshaffer:
This looks great

cedricziel:
Great!

tmatsuo:
FYI, it's deployed to everywhere in prod

Issue #130, opened 2016-07-22
[RFC] PHP 7.1


cedricziel:
[PHP 7.1 beta1 was released](http://php.net/archive/2016.php#id2016-07-21-1) :tada: 

Would you accept a pull request with a third runtime version?

tmatsuo:
Yes! We moved the test run to CircleCI a while ago. Let me check if tests for PRs from other repos just work.

tmatsuo:
Yeah it worked. Feel free to send us a PR :)

tmatsuo:
I hope we will have PHP 7.1 soon. Related #173

Issue #34, opened 2015-12-18
Pre-compile the PHP binaries


tmatsuo:
It is too slow to compile multiple PHP versions for each Docker build. Can we pre-compile the PHP binaries, upload and download from GCS or somewhere?

tmatsuo:
@sjlangley What do you think?

tmatsuo:
Probably we can use a gcs bucket and upload the PHP binary after test success upon the first compile, and download the binary for all the following build.

sjlangley:
is it a big deal? you only build them once then store the image right?

tmatsuo:
During the application development, it is not a problem at all. The issue here is the development time of this image itself. For example, even when just adding one single new extension, the test runs all of the build, which slows down the development process of this image itself.

If we put the binary somewhere, we can just download it and compile the new extension alone, which will be way faster.

cedricziel:
I'd appreciate binaries as well. The most important facette of this solution for would be that you could potentially switch out/pin patch level versions without having to compile software for an hour upfront.

PR #132, opened 2016-07-25
[WIP] Support for PHP 7.1

cedricziel:
Referring to #130 

Currently this uses PHP7 beta 1.

Integrity check is disabled for the time being.

The following is a short list of key points that came to my mind. I'm pretty sure I missed at least a few.
- [x] make sure 7.1 actually compiles
- [x] apply `parse_str` patch
- [ ] only use 7.1 if explicitly demanded, since it's beta software. May change when 7.1 actually gets released
- [ ] add PHP 7.1 specific test suite
- [ ] find out which keys are used for the 7.1 releases, add them,  re-enable integrity checks

Any help or comment is greatly appreciated and you may seamlessley send a pull-request to the `php71` branch on my fork and I'm happy to discuss the topic in general :)


cedricziel:
It seems I can't investigate why the build fails, as I can only access the first 4 MB of the the log file. I'll run a build locally to see what's fishy there. You may want to look if there's a way for me to get access to the rest of the log.

cedricziel:
Local build succeeds though. I' raised the timeout to 1hr in the circle ci configuration. Let's see what happens.

php-nginx/build-scripts/build_php71.sh
```diff
@@ -0,0 +1,150 @@
...
+
+# APC compatibility layer for APCu
+mkdir -p ${PHP_SRC}/ext/apcu-bc
+curl -SL "https://pecl.php.net/get/apcu_bc-1.0.3.tgz" -o apcu-bc.tar.gz
+tar -zxf apcu-bc.tar.gz -C ${PHP_SRC}/ext/apcu-bc --strip-components=1
+rm apcu-bc.tar.gz
+
+pushd ${PHP_SRC}
+# Consider adding an equivalent patch in later stages
+# patch -p1 < /build-scripts/php7-parse_str_harden.patch
```
tmatsuo: Did the patch fail?

circle.yml
```diff
@@ -18,4 +18,5 @@ dependencies:
 
 test:
   override:
-    - scripts/run_test_suite.sh
+    - scripts/run_test_suite.sh:
+        timeout: 3600
```
tmatsuo: Now the last build took 44 minutes. Is it safer to raise more?

php-nginx/build-scripts/build_php71.sh
cedricziel: I don't know - I wanted a vanilla build first, so I can go from a successful build, rather than a failing one. ;)

circle.yml
cedricziel: Circle has a maximum timeout of 2 hrs, I'll give it another 30 minutes on the next push.

php-nginx/build-scripts/build_php71.sh
tmatsuo: If you don't mind, can you try the patch?

cedricziel: Yep, did so yesterday - unfortunately the sources have diverged. - I guess I'll have to replicate it, but my C is .. a little bit rusty, so I postponed it :)

cedricziel: A note on this, as I crafted a new patch: I think calling `parse_str` without the second parameter has been deprecated and is subject to removal. In the future. https://wiki.php.net/rfc/deprecations_php_7_1#suggested_deprecations

tmatsuo:
@cedricziel Thanks for the pull request, and your work so far. We actually decided to change the way we compile php versions. We'll create deb package beforehand, and use them on demand. Here is the rationale. When we only support 2 major versions, the build time was too long. Related #34

cedricziel:
No problem - we have thought about such things beforehand and I'm fine with it. Thx for letting me know.

PR #133, opened 2016-07-26
Building the images with gcloud if credentials are available

tmatsuo:


scripts/build_images.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/usr/bin/env bash
```
bshaffer: I'm not a bash guy... this is more standard right?

scripts/build_images.sh
```diff
@@ -15,11 +15,44 @@
...
+        gcloud docker pull "${FULL_TAG}"
+        docker tag "${FULL_TAG}" "${IMAGE}"
+    else
+        # No credentials. Use local docker.
+        docker build -t "${IMAGE}" "${DIR}"
+    fi
+}
+
+build_image php-nginx php-nginx
+build_iamge php56 testapps/php56
```
bshaffer: `build_iamge`!!!

scripts/build_images.sh
```diff
@@ -15,11 +15,44 @@
...
 
-docker build -t php-nginx php-nginx
-docker build -t php56 testapps/php56
-docker build -t php56_custom testapps/php56_custom
-docker build -t php70 testapps/php70
-docker build -t php56_70 testapps/php56_70
-docker build -t php70_custom testapps/php70_custom
-docker build -t php56_nginx_conf testapps/php56_nginx_conf
-docker build -t php56_custom_configs testapps/php56_custom_configs
+TAG="rc"
```
bshaffer: release candidate? Maybe we should make this something more clear? This is not familiar to me in this context.

scripts/install_test_dependencies.sh
```diff
@@ -52,7 +57,7 @@ if [ "${CIRCLECI}" == "true" ]; then
     # They also overrides the PATH to use
     # /opt/google-cloud-sdk/bin/gcloud so we can not easily use our
     # own gcloud
-    sudo /opt/google-cloud-sdk/bin/gcloud -q components update gsutil
+    sudo /opt/google-cloud-sdk/bin/gcloud -q components update alpha app gsutil
 else
-    gcloud -q components update app
+    gcloud -q components update alpha app gsutil
```
bshaffer: Why `gsutil` here? Is it because `alpha` depends on it in some way?

bshaffer:
There's some typo fixes and a few misc questions, but otherwise this looks like a great change!

scripts/build_images.sh
tmatsuo: Yeah it's pretty normal.

scripts/build_images.sh
tmatsuo: Doh, fixed

scripts/install_test_dependencies.sh
tmatsuo: Let me remove this

scripts/build_images.sh
tmatsuo: Changed it to `test-build`

tmatsuo:
@bshaffer Thanks! I fixed several stuff. PTAL

cedricziel:
Looks good

scripts/build_images.sh
```diff
@@ -15,11 +15,51 @@
...
+    if [ -f "${PHP_DOCKER_GOOGLE_CREDENTIALS}" ]; then
+        # Build the image with container builder service if we have
+        # credentials.
+        FULL_TAG="gcr.io/${GOOGLE_PROJECT_ID}/${IMAGE}:${TAG}"
+        SRC_DIR="${SRC_TMP}/${DIR}"
+        mkdir -p $(dirname ${SRC_DIR})
+        cp -R "${DIR}" "${SRC_DIR}"
+        # Replace the FROM line to point to our image in gcr.io.
+        sed -i -e "s|FROM php-nginx|FROM ${BASE_IMAGE}|" "${SRC_DIR}/Dockerfile"
+        gcloud -q alpha container builds create "${SRC_DIR}" --tag "${FULL_TAG}"
```
bshaffer: possibly dumb question - Why do we need to move everything into a tmp dir?

tmatsuo: It's not strictly needed, but if you run this locally, it will rewrite the source code.

bshaffer: ahh, it's because of the `sed`. Got it, thanks.

bshaffer:
LGTMz!

tmatsuo:
Thx!

PR #134, opened 2016-08-01
Added `-f` option to `docker tag`

tmatsuo:
On environments like jenkins, or local workstations, `docker tag` will fail if there's always the tag.


bshaffer:
👍

Issue #135, opened 2016-08-02
Reintroduce scripts to composer install


bshaffer:
We added `--no-scripts` to the composer install commmand in https://github.com/GoogleCloudPlatform/php-docker/pull/69. Why did we do this?

This could be a very way for users to run simple pre-deployment tasks (such as changing file permissions, warming up caches, etc) without having to extent the `Dockerfile`. We can add `--no-interaction` to the composer install command, to ensure the build can complete.

@cedricziel

tmatsuo:
Yeah we discuss enabling running script with composer, because it is a handy way of doing pre-work before serving the app.

I thought you needed to disable script because symfony's script raises an error.

Can you provide a reproducible app so that we can investigate for workaround solutions?

bshaffer:
Symfony Standard uses composer's `post-install-cmd` to generate [`parameters.yml`](https://github.com/symfony/symfony-standard/blob/master/app/config/parameters.yml.dist#L1) (see [here](https://github.com/symfony/symfony-standard/blob/master/composer.json#L37)). Laravel uses `post-install-cmd` to run [`php artisan optimize`](https://github.com/laravel/laravel/blob/master/composer.json#L40).

cedricziel:
Long story short:

The `ONBUILD` hook executes at a point in time, where I can do nothing to interfere or fullfill package wise dependencies, such as installing debian packages or building php extensions.

Symfony for example compiles the dependency injection container and compiles proxy classes through the composer script handler which will ultimately result in errors if I can not supply all needed depencies or customize the image to my needs *before* the very first composer script execution.

If you think about enabling the scripts again, please provide a tag that doesnt run them.

To understand a more complex problem, I've written a bunch of comments about my initial intention: Removing the `--no-dev` argument in https://github.com/GoogleCloudPlatform/php-docker/pull/66
The default env with symfony is `dev`. I won't repeat all of that here, but in short, the solution would be to mimic the heroku images in that regard (see #66)

tmatsuo:
Ok, so how about this?

* During ONBUILD, run `composer install --no-scripts --no-dev --prefer-dist --optimize-autoloader --no-interaction --no-ansi --no-progress`
* in the `entrypoint.sh` (it's executed at runtime), run the composer again without `--no-scripts`, `composer install --no-dev --prefer-dist --optimize-autoloader --no-interaction --no-ansi --no-progress`

[...]

PR #136, opened 2016-08-02
adds "run-script post-deploy-cmd" to composer.sh and adds tests

bshaffer:


bshaffer:
see #135

tmatsuo:
Some scripts which needs modification to the base image will fail.

I think we need to run `composer install` 2 times as described at: https://github.com/GoogleCloudPlatform/php-docker/issues/135#issuecomment-237001254

bshaffer:
@tmatsuo @cedricziel I added support for running custom scripts in `entrypoint.sh` for `post-deploy-cmd`. PTAL.

php-nginx/entrypoint.sh
```diff
@@ -114,4 +114,16 @@ sed -i "s|%%DOC_ROOT%%|${DOCUMENT_ROOT}|g" "${PHP_DIR}/lib/php.ini"
...
 ${PHP_DIR}/bin/php /whitelist_functions.php
 
+# run the composer scripts for post-deploy
+COMPOSER_CMD="${PHP_DIR}/bin/php \
+    -d suhosin.executor.include.whitelist=phar \
+    -d suhosin.executor.func.blacklist=none \
+    -d disable_functions= \
+    -d memory_limit=-1 \
+    -d max_input_time=-1 \
+    /usr/local/bin/composer"
```
tmatsuo: `--no-ansi` ?

testapps/php56_custom/TestScript.php
```diff
@@ -0,0 +1,37 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+namespace PhpDocker\TestApps\Php56Custom;
+
+use Symfony\Component\ClassLoader\ClassCollectionLoader;
+use Symfony\Component\Filesystem\Filesystem;
+use Symfony\Component\Process\Process;
+use Symfony\Component\Process\PhpExecutableFinder;
```
tmatsuo: Are they used?

tmatsuo:
LGTM

But let's wait for @cedricziel for his opinion.

This change won't affect your stuff because we're using custom script name.

Then we can say "To customize further, add a composer script named 'post-deploy-cmd'" in our guides.

How does it sound?

cedricziel:
Sounds good! :+1:

tmatsuo:
@bshaffer 
I think it's ready. Can I merge this? I will create another PR for running composer with `www-data` (or with the same uid)

bshaffer:
@tmatsuo okay, sounds good. This script should still be run as root though!

PR #138, opened 2016-08-09
Run composer as www-data.

tmatsuo:
Under the cover it does many things:

* Temporarily enable shell for `www-data` user
* Run composer as `www-data`
* chown -R www-data.www-data ${APP_DIR}
* Permission changes regarding to this
* After running the `post-deploy-cmd`, disable the shell entry for `www-data` user and lock down the ${DOCUMENT_ROOT} (instead of locking down the entire ${APP_DIR})

However, looking at the surface, it is very simple:
* For simple customization, you can use `post-deploy-cmd` which is executed by `www-data`
* ${DOCUMENT_ROOT} is locked down to `root.www-data` 0550, but ${APP_DIR} is owned by `www-data` and it should be writable while it is still recommended to use `/tmp`
* Because www-data runs `composer install`, the files generated by the composer scripts can be accessible from the web app

My only concern is https://github.com/docker/docker/issues/6047 which may cause nasty permission errors.

### DIFF
--- a/php-nginx/Dockerfile
+++ b/php-nginx/Dockerfile
@@ -92,16 +92,15 @@ COPY logrotate.app_engine /etc/logrotate.d/app_engine
 
 COPY entrypoint.sh composer.sh whitelist_functions.php /
 
-RUN chgrp www-data composer.sh && chmod +x /entrypoint.sh /composer.sh
-
-# Temporary enable the shell for www-data
-# We will disable it in entrypoint.sh
-RUN chsh -s /bin/bash www-data
+RUN chgrp www-data composer.sh && chmod +x /entrypoint.sh /composer.sh \
+    # Temporary enable the shell for www-data
+    # We will disable it in entrypoint.sh
+    && chsh -s /bin/bash www-data
 
 # A script for extracting PHP version from composer.json.
 COPY detect_php_version.php /tmp/detect_php_version.php
-RUN chgrp www-data /tmp/detect_php_version.php
-RUN cd /tmp && su www-data -c "${PHP_DIR}/bin/php \
+RUN chgrp www-data /tmp/detect_php_version.php && cd /tmp \
+    && su www-data -c "${PHP_DIR}/bin/php \
         -d suhosin.executor.include.whitelist=phar \
         -d suhosin.executor.func.blacklist=none \
         -d memory_limit=-1 \
--- a/php-nginx/Dockerfile
+++ b/php-nginx/Dockerfile
@@ -55,7 +55,8 @@ ENV NGINX_DIR=/opt/nginx \
     NGINX_VERSION=1.10.1 \
     PHP56_VERSION=5.6.24 \
     PHP70_VERSION=7.0.9 \
-    PATH=/opt/php/bin:$PATH
+    PATH=/opt/php/bin:$PATH \
+    WWW_HOME=/var/www
 
 # gpgkeys for verifying the tarballs
 COPY gpgkeys /gpgkeys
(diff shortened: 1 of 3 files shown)


php-nginx/Dockerfile
```diff
@@ -91,28 +92,30 @@ COPY logrotate.app_engine /etc/logrotate.d/app_engine
 
 COPY entrypoint.sh composer.sh whitelist_functions.php /
 
-RUN chmod +x /entrypoint.sh /composer.sh
+RUN chgrp www-data composer.sh && chmod +x /entrypoint.sh /composer.sh
+
+# Temporary enable the shell for www-data
+# We will disable it in entrypoint.sh
+RUN chsh -s /bin/bash www-data
```
bshaffer: Do we want to combine this with the `RUN` line above?

php-nginx/Dockerfile
```diff
@@ -91,28 +92,30 @@ COPY logrotate.app_engine /etc/logrotate.d/app_engine
...
 
 # A script for extracting PHP version from composer.json.
 COPY detect_php_version.php /tmp/detect_php_version.php
-RUN cd /tmp && ${PHP_DIR}/bin/php \
+RUN chgrp www-data /tmp/detect_php_version.php
+RUN cd /tmp && su www-data -c "${PHP_DIR}/bin/php \
```
bshaffer: Should this be combined with the `RUN` line above?

tmatsuo: Not a big deal, but I can do that

php-nginx/Dockerfile
tmatsuo: I just wanted to add a comment

php-nginx/composer.sh
```diff
@@ -66,9 +67,13 @@ EOF
     # no need for the token to stay around in the env
     unset COMPOSER_GITHUB_OAUTH_TOKEN
 
+    # Workaround for https://github.com/docker/docker/issues/6047
+    # We want to remove when Container Builder starts to use newer Docker.
+    rm -rf ${APP_DIR}/vendor
```
bshaffer: I really don't like force-removing the vendor directory. When will Container Builder be updated?

tmatsuo: Not sure, but currently this is the easiest workaround

php-nginx/entrypoint.sh
```diff
@@ -114,17 +114,28 @@ sed -i "s|%%DOC_ROOT%%|${DOCUMENT_ROOT}|g" "${PHP_DIR}/lib/php.ini"
...
     -d suhosin.executor.include.whitelist=phar \
     -d suhosin.executor.func.blacklist=none \
     -d disable_functions= \
     -d memory_limit=-1 \
     -d max_input_time=-1 \
     /usr/local/bin/composer \
     --no-ansi"
-if $COMPOSER_CMD run-script -l | grep -q "post-deploy-cmd"; then
-    $COMPOSER_CMD run-script --no-interaction post-deploy-cmd
+    if su www-data -c "${COMPOSER_CMD} run-script -l" \
```
bshaffer: I am worried we may need to run these as root down the line, but I don't mind making this restrictive to start.

php-nginx/composer.sh
```diff
@@ -23,10 +23,10 @@ DEFAULT_PHP_VERSION="5.6"
 if [ -f ${APP_DIR}/composer.json ]; then
     # Extract php version from the composer.json.
     CMD="${PHP_DIR}/bin/php /tmp/detect_php_version.php ${APP_DIR}/composer.json"
-    PHP_VERSION=`${CMD}`
+    PHP_VERSION=`su www-data -c "${CMD}"`
 
     # Remove the vendor directory for the temporary script above.
-    rm -rf /tmp/vendor
+    rm -rf /tmp/vendor /tmp/detect_php_version.php /tmp/composer.*
```
bshaffer: where does the tmp composer come from? does the script output it? If so, we should have the script clean it up.

php-nginx/entrypoint.sh
tmatsuo: Is it for `apt-get` or something? If so, I have a plan for allowing apt-get for `www-data` user.

php-nginx/composer.sh
tmatsuo: It's done in `Dockerfile`

tmatsuo:
@cedricziel FYI

tmatsuo:
@bshaffer PTAL!

php-nginx/composer.sh
bshaffer: Ok. Can we update the comment to reflect that? `# Remove files and directories created in Dockerfile` or something along those lines

bshaffer:
One nit comment fix, otherwise LGTM

tmatsuo:
@bshaffer Thanks!
I will merge this once the test passes.

cedricziel:
Thx for the mentionm @tmatsuo. I'll try it, once I find time for it :)

PR #139, opened 2016-08-10
Provide a default nginx-app.conf

tmatsuo:
Note: with this image, we can just remove the `nginx-app.conf` in our most of the (probably all?) sample applications.

### DIFF
--- a/php-nginx/README.md
+++ b/php-nginx/README.md
@@ -112,28 +112,26 @@ runtime_config:
 
 ## How to change nginx.config
 
-Put `nginx-app.conf` file which includes piece of nginx configuration
-for the `server` section in the project top directory. Then it will be
-moved to an appropriate directory and included from the main nginx
-configuration file.
+We put `nginx-app.conf` file which is included in the `server` section
+of the main configuration file. Here is the default configuration file
+which will likely just work with most of the modern web frameworks in
+simple cases.
 
-Assumption is that you don't often need to entirely modify the
-[default nginx file](nginx.conf).
-There is an `include` directive in the `server` section and your
-configuration file will be included there.
-
-Here is an example configuration for silex.
-
-nginx-app.conf:
+The default nginx-app.conf:
 
 ```ini
 location / {
-  # try to serve file directly, fallback to front controller
+  # try to serve files directly, fallback to the front controller
   try_files $uri /index.php$is_args$args;
 }
 ```
 
-I hope this mechanism can cover most of the web frameworks, but let us
+You might need to add some rewrite rules, or you might want to change
+the behavior. In those cases, put a file named `nginx-app.conf` in the
+project root directory. Then the runtime will override the default
+file with the file you provided.
+
+I hope this mechanism can cover most of your use cases, but let us
 know if you found otherwise.
 
 ## Change the PHP version
--- a/php-nginx/README.md
+++ b/php-nginx/README.md
@@ -112,12 +112,9 @@ runtime_config:
 
 ## How to change nginx.config
 
-We put `nginx-app.conf` file which is included in the `server` section
-of the main configuration file. Here is the default configuration file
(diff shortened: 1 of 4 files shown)


php-nginx/README.md
```diff
@@ -112,28 +112,26 @@ runtime_config:
...
 location / {
-  # try to serve file directly, fallback to front controller
+  # try to serve files directly, fallback to the front controller
   try_files $uri /index.php$is_args$args;
 }
 ```
 
-I hope this mechanism can cover most of the web frameworks, but let us
+You might need to add some rewrite rules, or you might want to change
+the behavior. In those cases, put a file named `nginx-app.conf` in the
```
bshaffer: Symfony uses `app.php` in the root of the project as a convention, so you should also add instructions on how to do this:

> In this case, `index.php` is used as the framework front controller. You may need to change this to something different for your project. The Symfony framework, for instance, uses `app.php` instead of `index.php`.

php-nginx/README.md
```diff
@@ -112,28 +112,26 @@ runtime_config:
 
 ## How to change nginx.config
 
-Put `nginx-app.conf` file which includes piece of nginx configuration
-for the `server` section in the project top directory. Then it will be
-moved to an appropriate directory and included from the main nginx
-configuration file.
+We put `nginx-app.conf` file which is included in the `server` section
```
bshaffer: This is hard to read. How about this?

> An `nginx-app.conf` configuration file is included in the `server` section of the main configuration file. The default configuration file looks like this:
> //...
>  To define a custom configuration file, put a file named `nginx-app.conf` in the project root directory. The runtime will override the default file with the file you provided.

php-nginx/README.md
bshaffer: I added this to the suggested edit above.

php-nginx/README.md
tmatsuo: Done!

bshaffer:
It would be great to include a section with the proper `app.yaml` configuration and (if necessary) `nginx-app.conf` file for popular frameworks, i.e.

Framework | Document root | front controller
------------ | ------------- | ---
Symfony2 / Symfony 3 | `web` | `app.php`
Drupal8| project root | `index.php`
Laravel | `public` | `index.php`
Wordpress | project root | [custom `nginx-app.conf`](https://github.com/GoogleCloudPlatform/php-docs-samples/blob/master/appengine/wordpress/src/files/flexible/nginx-app.conf)

tmatsuo:
@bshaffer Great suggestion, but I want put those info on our documentation, not on the `README` in this repo.

tmatsuo:
We exhausted quotas for Container Builder API. Probably I've worked too much today!

cedricziel:
+1 for adding it to the documentation. Great an necessary addition :+1:

tmatsuo:
Ok, I'll add the section, but can we do that in another PR?

tmatsuo:
@bshaffer PTAL!

bshaffer:
LGTM!

tmatsuo:
Alright, the current image is available at:
gcr.io/google_appengine/php:2016-08-10_18_55
and
gcr.io/google_appengine/php:staging

Please feel free to test it.

You may still get a nasty permission error if you try to modify files in sub directories in ${APP_DIR} due to https://github.com/docker/docker/issues/6047
(I hope it will be fixed when we move newer docker on container builder)

Possible wordarounds:
- use "--storage-driver=devicemapper" for docker storage driver and build the image locally
- use `/tmp` for temporary files

tmatsuo:
@bshaffer tried that image and gave good feedback. We should loosen suhosin etc for php cli. I'll work on that next and I'll stage the image again, after doing do.

tmatsuo:
See also #141

PR #142, opened 2016-08-11
Added php-cli.ini for easy cli access.

### LINKED ISSUE #141: Loosen suhosin etc for php-cli

tmatsuo:
fixes #141

* added `php-cli.ini` with all the restrictions removed so that php cli can call any functions.
* create a symlink at `/usr/bin/php` so that you can just use `php` everywhere
* put a stricter `php-cli.ini` before starting the web app


php-nginx/build-scripts/build_php56.sh
```diff
@@ -80,7 +80,7 @@ rm -f configure
     --enable-gd-native-ttf \
     --enable-intl=shared \
     --enable-mailparse \
-    --enable-mbstring=shared \
+    --enable-mbstring \
```
tmatsuo: There is no shared flag

php-nginx/build-scripts/build_php56.sh
```diff
@@ -104,7 +104,7 @@ rm -f configure
     --with-xsl=shared \
     --with-fpm-user=www-data \
     --with-fpm-group=www-data \
-    --with-mysql=mysqlnd \
+    --with-mysql \
```
tmatsuo: `=mysqlnd` is not required

tmatsuo:
The image from this PR is pushed as
gcr.io/google_appengine/php:2016-08-11_06_52

php-nginx/entrypoint.sh
```diff
@@ -138,4 +134,7 @@ chmod -R 550 ${DOCUMENT_ROOT}
 # Change the www-data's shell back to /usr/sbin/nologin
 chsh -s /usr/sbin/nologin www-data
 
+# Remove the loose php-cli.ini
+rm ${PHP_DIR}/lib/php-cli.ini
```
bshaffer: Why do we remove the `php-cli.ini` file? It would be useful for debugging purposes.

tmatsuo: If you enable `exec` or the sort, it will open a door to run the loose php from web app. It can be an easy route for intruders.

bshaffer: but we're locking down `exec` specifically so this doesn't happen. Until we can actually output a proper error response on the CLI, we really can't keep suhosin activated.

tmatsuo: You can enable `exec` with `runtime:php`. This means it should be still as secure as possible even when you enable `exec`.

If it's for easier debugging, I can provide an alternate easy way later without loosen the security of the runtime.

bshaffer: I think we can make the error message `PHP Warning: function() has been disabled for security reasons`

tmatsuo: Seems like a nice idea, can you file an issue?

bshaffer: We can set `display_errors=STDERR`, which will fix this issue. I'm looking into it to see if we want to set it for web as well, since STDERR should just be logged.  I verified this locally:
```php
// test.php
<?php
exec('php -r "phpinfo();"', $output, $return_var);
var_dump($output, $return_var);
// array(0) { } int(0)
```
When accessed through a web server, the CLI error message is not displayed, only if it's run on the CLI (through `stderr`)

tmatsuo: Seems we should consider using that, but I want to make that change in different PR

bshaffer: How about we add a `php-cli.ini` with `display_errors=STDERR`, and then I'll be happy. This means nothing changes for web configuration, or for scripts ran with `exec`, but will output helpful messages for users running scripts on the CLI.

php-nginx/entrypoint.sh
```diff
@@ -116,18 +116,14 @@ ${PHP_DIR}/bin/php /whitelist_functions.php
 
 if [ -f "${APP_DIR}/composer.json" ]; then
     # run the composer scripts for post-deploy
-    COMPOSER_CMD="${PHP_DIR}/bin/php \
-    -d suhosin.executor.include.whitelist=phar \
-    -d suhosin.executor.func.blacklist=none \
-    -d disable_functions= \
-    -d memory_limit=-1 \
-    -d max_input_time=-1 \
+    COMPOSER_CMD="env PATH=${PATH}:${PHP_DIR}/bin ${PHP_DIR}/bin/php \
```
bshaffer: this seems like the wrong place to declare our PATH var

tmatsuo: Somehow other means don'e work for us. Patch welcome :)

php-nginx/entrypoint.sh
tmatsuo: Done

tmatsuo: PTAL

tmatsuo: FYI I tested running composer with suhosin and `display_errors=STDERR`, but it didn't display anything.

tmatsuo: @bshaffer 

Basically, adding `display_errors=STDERR` doesn't improve debugging experience with suhosin enabled.

I would suggest that we rollback the last commit and provide a script for better debugging in another PR. WDYT?

bshaffer: @tmatsuo I tested this on the container running locally and it worked. I wonder if there's something else configured differently.

tmatsuo: @bshaffer Ok, if it works, that's good. Maybe I will retry.

tmatsuo: @bshaffer Yeah, it works for me too.

tmatsuo:
@bshaffer Anything left?

tmatsuo:
@bshaffer OK, just added a symlink at `/usr/bin/php` for easy access when we use `su`. PTAL

php-nginx/entrypoint.sh
```diff
@@ -116,9 +116,7 @@ ${PHP_DIR}/bin/php /whitelist_functions.php
 
 if [ -f "${APP_DIR}/composer.json" ]; then
     # run the composer scripts for post-deploy
-    COMPOSER_CMD="env PATH=${PATH}:${PHP_DIR}/bin ${PHP_DIR}/bin/php \
-    /usr/local/bin/composer \
-    --no-ansi"
+    COMPOSER_CMD="php /usr/local/bin/composer --no-ansi"
```
bshaffer: nit: I think we can just call this directly w/o setting `COMPOSER_CMD` now, i.e. 

```sh
if su www-data -c php /usr/local/bin/composer --no-ansi run-script -l" \
# ...
```

bshaffer:
one nit, otherwise lgtm!

php-nginx/entrypoint.sh
tmatsuo: We use ${COMPOSER_CMD} twice and I think it's fine to keep the variable.

tmatsuo:
@bshaffer thanks!
I'm going to merge this. If you still don't like ${COMPOSER_CMD}, please open another PR.

tmatsuo:
FYI, the new image is staged at the following locations:
gcr.io/google_appengine/php:2016-08-18_17_03
gcr.io/google_appengine/php:staging

tmatsuo:
Also pushed to `gcr.io/google_appengine/php:latest`

PR #144, opened 2016-08-18
removes COMPOSER_CMD

bshaffer:
addendum to #142 - misc code cleanup

### DIFF
--- a/php-nginx/entrypoint.sh
+++ b/php-nginx/entrypoint.sh
@@ -116,11 +116,12 @@ ${PHP_DIR}/bin/php /whitelist_functions.php
 
 if [ -f "${APP_DIR}/composer.json" ]; then
     # run the composer scripts for post-deploy
-    COMPOSER_CMD="php /usr/local/bin/composer --no-ansi"
-    if su www-data -c "${COMPOSER_CMD} run-script -l" \
+    if su www-data -c "php /usr/local/bin/composer --no-ansi run-script -l" \
         | grep -q "post-deploy-cmd"; then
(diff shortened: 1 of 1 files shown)


tmatsuo:
I'm just curious. Why this is better?

bshaffer:
The only reason for `$COMPOSER_CMD` was because the command was so long at the time (spanning ~10 lines) that it was more readable to set a variable. This is no longer the case, so removing a level of indirection (i.e. setting a variable and calling the variable) is more readable.

tmatsuo:
There is still another trivial but good reason for having a variable. It's used twice. I wouldn't mind keeping the variable.

bshaffer:
Then we should use it twice [here](https://github.com/GoogleCloudPlatform/php-docker/blob/master/php-nginx/composer.sh#L56), and set `${APP_DIR}/composer.json` to `$COMPOSER_JSON`. I don't think using something twice is a good reason to set a variable. The trade of indirection for code reuse is not always worth it.

tmatsuo:
Those 2 calls and the assignment are very close in the script. It's not difficult to read at all.

bshaffer:
It's not hard to read as is, but it's easier to read without it. Someone scanning the file doesn't know what `COMPOSER_CMD` is.  It's a nit, but this way is cleaner. Why have a variable you don't need?

tmatsuo:
LGTM let's merge this after the test passes.

bshaffer:
Thanks! I know it's a nit, but I am glad we merged it!

PR #146, opened 2016-09-16
Added USER directive to run the app as non-root user 

tmatsuo:
- Abandoned running cron
- It runs the php56_custom app in the testapp dir correctly only with devicemapper storage backend rather than the default aufs. So it is possible to fail when building with Cloud Container Builder. Container Builder is using aufs.
- It used to lock down the permission of ${DOCUMENT_ROOT} directory after running the custom script; like root.www-data 0550. Now we can not do that.
- I needed to loosen permissions of some directories to keep features dynamically done in `entrypoint.sh`

The command for testing the image locally is:

```
 $ docker run -t -i -u 1000 -p 127.0.0.1:8080:8080 php56_custom
```


tmatsuo:
The tests are green.

We need answer some questions below in order to merge this change.

- Are we OK with the looser file permissions?
- Who is responsible with log rotation?

tmatsuo:
Originally the image change the owner of the files under ${DOCUMENT_ROOT} to root.www-data with 0550 permissions. It is done in the `entrypoint.sh` after we run custom composer script.

Now the `entrypoint.sh` is executed by `www-data`, so we can not change the permission as before. The commit 58abe19 is an attempt to achieve this with `sudo`.

tmatsuo:
Maybe we should not make the permission too strict. If we do root.www-data 0550, it is very hard for users to ssh into the VMs and debug.

Probably we can just change the file permission to 0500 and owned by the uid, so that users can override the file.

tmatsuo:
Closing for now

PR #150, opened 2016-09-28
Switch base image from base to debian8

sharifelgamal:
The newest debian images for jessie (gcr.io/google-appengine/debian8:latest) now has all the extra env variables and ca-certs that were in base.

cc @dlorenc


tmatsuo:
Thanks, I'll merge this once the test passes

tmatsuo:
Released to prod

PR #155, opened 2016-10-19
Initial split of php-base image

deustis:
@tmatsuo @dlorenc this is an attempt to split the basic PHP install from nginx and GAE-specific installations.


tmatsuo:
What's the size of the final image? The current version (without this patch), it's 458.9MB.

deustis:
When I build this image locally I get:
- php-base: 394.5 MB
- php-nginx: 469.7 MB

Still working on the README.

PR #163, opened 2016-11-17
Drop access to some files by default.

tmatsuo:


bshaffer:
What is the justification for this?

php-nginx/nginx.conf
```diff
@@ -78,5 +78,10 @@ http {
 
         # Allow users to specify their own configuration
         include /etc/nginx/conf.d/*.conf;
+
+        # Drop accesses for some files
+        location = /favicon.ico { access_log off; log_not_found off; }
+        location ~ /\.          { access_log off; log_not_found off; deny all; }
+        location ~ ~$           { access_log off; log_not_found off; deny all; }
```
bshaffer: It would be great to have this added to `.dockerignore` at some point

tmatsuo: noted!

PR #165, opened 2016-12-07
Fixed the build on our jenkins

tmatsuo:


scripts/run_test_suite.sh
```diff
@@ -20,7 +20,9 @@ set -ex
 
 # Run php-cs-fixer.
 # We want to fail fast for coding standard violations.
-vendor/bin/php-cs-fixer fix --dry-run --diff .
+if [ -z "${SKIP_CS_CHECK}" ]; then
+    vendor/bin/php-cs-fixer fix --dry-run --diff .
```
bshaffer: nit: we should remove the `.` when we upgrade `.php_cs` because then the filesystem is determined in the configuration file.

tmatsuo: Done

tmatsuo:
@bshaffer Thanks. I needed to migrate the e2e test into the cloudbuild step. PTAL

tmatsuo:
Alright, jenkins build is green now :)

cloudbuild-test-runner/Dockerfile
```diff
@@ -14,22 +14,30 @@
...
-    && php -r "copy('https://getcomposer.org/installer', 'composer-setup.php');" \
-    && php composer-setup.php --filename=composer --install-dir=/opt/bin \
-    && php -r "unlink('composer-setup.php');" \
-    && wget https://phar.phpunit.de/phpunit.phar \
-    && chmod +x phpunit.phar \
-    && mv phpunit.phar /opt/bin/phpunit \
-    && apk del wget \
-    && rm -rf /var/cache/apk/*
+    ca-certificates coreutils unzip --update && \
+    php -r "copy('https://getcomposer.org/installer', 'composer-setup.php');" && \
```
bshaffer: Do we want to add the following for security?

```sh
php -r "if (hash_file('SHA384', 'composer-setup.php') === rtrim(file_get_contents('https://composer.github.io/installer.sig'))) { echo 'Installer verified'; } else { echo 'Installer corrupt'; unlink('composer-setup.php'); } echo PHP_EOL;"

tmatsuo: Yes will do

tmatsuo:
@bshaffer Now we verify the composer installer. PTAL

tmatsuo:
The last failure was caused because the cloudbuild hit the GCS rate limitting for log injestion.
It happens when we run multiple builds (on circleci, and jenkins at the same time, etc)

Workarounds are
- to lower log verbosity
- to use different projects on different CI

PR #166, opened 2016-12-09
Stopped removing the vendor dir in composer.sh

tmatsuo:

PR #169, opened 2016-12-13
Revert "Stopped removing the vendor dir in composer.sh"

tmatsuo:
Reverts GoogleCloudPlatform/php-docker#166


tmatsuo:
See https://github.com/GoogleCloudPlatform/php-docker/pull/166#issuecomment-266646326 for the reason for the revert

PR #171, opened 2016-12-13
Version upgrades

tmatsuo:
`--enable-memcached-sasl` option is actually not needed because the default is enabled. Now it's just installing from pecl which should be fine. It will also remove the warnings about missing so files at startup.

### DIFF
--- a/php-nginx/build-scripts/build_php56.sh
+++ b/php-nginx/build-scripts/build_php56.sh
@@ -34,11 +34,6 @@ tar -zxf php.tar.gz -C ${PHP_SRC} --strip-components=1
 rm php.tar.gz
 rm php.tar.gz.asc
 
-mkdir -p ${PHP_SRC}/ext/memcached
-curl -SL "http://pecl.php.net/get/memcached" -o memcached.tar.gz
-tar -zxf memcached.tar.gz -C ${PHP_SRC}/ext/memcached --strip-components=1
-rm memcached.tar.gz
-
 rm -rf ${PHP_SRC}/ext/json
 mkdir -p ${PHP_SRC}/ext/json
 curl -SL "https://pecl.php.net/get/jsonc" -o jsonc.tar.gz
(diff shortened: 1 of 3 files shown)


php-nginx/build-scripts/build_php56.sh
```diff
@@ -128,9 +121,10 @@ mkdir -p ${PHP56_DIR}/lib/conf.d
 
 # Install shared extensions
 ${PHP56_DIR}/bin/pecl install memcache
```
bshaffer: Why do we install `memcache`? Memcached should suffice.

bshaffer:
From [the memcached documentation](http://php.net/manual/en/memcached.installation.php):

> SASL authentication support is disabled by default. To enable it, use --enable-memcached-sasl switch. This requires that libsasl2 has been installed and that libmemcached has been built with SASL support enabled.

tmatsuo:
@bshaffer 

Yeah it's confusing, but as I looked at the source repo of that extension, it's enabled by default and the build log says

```
Step #0: checking whether to enable memcached sasl support... yes
```

php-nginx/build-scripts/build_php56.sh
tmatsuo: It's just installed as shared extension and it's not enabled by default, so I'd keep it.

There are still some code depending on `Memcache` class out there.

tmatsuo:
@bshaffer Also I'll put the new image as `staging` and manually test the sasl auth (for now) with our samples. It's manual now, but we'll want to come up with an easy way to run our samples against the image in `staging`.

tmatsuo:
@bshaffer 

I confirmed that sasl support is enabled by default by spotting the commit:
https://github.com/php-memcached-dev/php-memcached/commit/ace57718002ed7d30b620694fbe68e76abc821fd

which is marked as `2.2.0` which is released in 2014

tmatsuo:
@bshaffer A documentation bug reported: https://bugs.php.net/bug.php?id=73736

php-nginx/build-scripts/build_php56.sh
bshaffer: I think we should remove it before this goes beta, as it may be confused with our (forthcoming) Memcached support.

tmatsuo:
@bshaffer Thanks!

tmatsuo:
JFYI I tested the image manually and our memcached sample works with redislab

PR #177, opened 2017-01-05
Use memcached and jsonc debian packages

tmatsuo:
Use deb packages generated by #178


tmatsuo:
@dazuma @bshaffer Can you take a look?

tmatsuo:
I'm pretty certain that we need the binaries built by #180 for the test to pass.

PR #180, opened 2017-01-07
Added /app to the --with-config-file-scan-dir option.

tmatsuo:
This is intentionally not using the environment variable.

To change the `APP_DIR` env var, first you need to change it in our PHP
binaries, upload those packages, then change it in
`php-nginx/Dockerfile`.

### DIFF
--- a/deb-package-builder/debian/rules.in
+++ b/deb-package-builder/debian/rules.in
@@ -12,7 +12,7 @@ override_dh_auto_configure:
 	&& ./buildconf --force \
 	&& dh_auto_configure -- --prefix=/opt/php${SHORT_VERSION} \
 	--with-config-file-path=/opt/php${SHORT_VERSION}/lib \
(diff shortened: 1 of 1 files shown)


tmatsuo:
@bshaffer 

Can you take a look? It turned out that we need to add /app to the config scan dir anyways, especially when we run composer install for users.

tmatsuo:
We run `composer install` way before we copy the file. With the current design, we need to delay copying because we need to refer to the environment variable to determine which file to copy. When we run `composer install`, it is better to have `/app` as a scan dir.

I think we can do better when we move to the builder model, but now I think we need this

tmatsuo:
Also, the original build script used to have ${APP_DIR} as the scan dir, so it's actually safe to add

bshaffer:
But this means other `*.ini` files in that directory will be parsed as php config files, but we should only parse `php.ini`.

tmatsuo:
@bshaffer 

Again, this is an old behavior and needed to migrate to debian package model. We can fix that sub optimal situation when we moved to the builder model

PR #182, opened 2017-01-09
Building more extensions

tmatsuo:
Except for grpc due to https://github.com/grpc/grpc/issues/9281


tmatsuo:
@dazuma Can you take a look?

cc @bshaffer

tmatsuo:
I found a workaround for grpc, and added it too.

deb-package-builder/extensions/grpc/grpc-fix-path.patch
```diff
@@ -0,0 +1,16 @@
...
+@@ -5,7 +5,7 @@ if test "$PHP_GRPC" != "no"; then
+   dnl Write more examples of tests here...
+ 
+   dnl # --with-grpc -> add include path
+-  PHP_ADD_INCLUDE(../../grpc/include)
+-  PHP_ADD_INCLUDE(../../grpc/src/php/ext/grpc)
+-  PHP_ADD_INCLUDE(../../grpc/third_party/boringssl/include)
++  PHP_ADD_INCLUDE(./include)
++  PHP_ADD_INCLUDE(./src/php/ext/grpc)
++  PHP_ADD_INCLUDE(./third_party/boringssl/include)
```
bshaffer: Should this be a PR to `grpc/grpc`? It's strange because neither path exists, but the updated path at least exists when the file is copied to the root of the library, so it's a step up.

tmatsuo: I propose it at https://github.com/grpc/grpc/issues/9281 and there seems to be another issue with `pecl install`

deb-package-builder/extensions/suhosin/debian/ext-suhosin.ini
```diff
@@ -0,0 +1,9 @@
+extension=suhosin.so
+
+; Configure the suhosin defaults that we want.
+suhosin.executor.disable_emodifier = On
+suhosin.executor.func.blacklist="escapeshellarg, escapeshellcmd, exec, highlight_file, lchgrp, lchown, link, symlink, passthru, pclose, popen, proc_close, proc_get_status, proc_nice, proc_open, proc_terminate, shell_exec, show_source, system, gc_collect_cycles, gc_enable, gc_disable, gc_enabled, getmypid, getmyuid, getmygid, getrusage, getmyinode, get_current_user, phpinfo, phpversion, php_uname"
+suhosin.log.file = 511
+suhosin.log.file.name = /var/log/app_engine/app-suhosin.log
```
tmatsuo: @bshaffer Also, the logs should be available on Cloud Logging too.

PR #184, opened 2017-01-10
Use more debian packages from our cloud-apt repo

tmatsuo:
Backward incompatible changes:
* memcache is no more available for PHP 7.0.x (memcached is available)
* suhosin is enabled just before running supervisord
* suhosin is under the simulation mode by default


tmatsuo:
@bshaffer First try for migrating to the debian packages.

tmatsuo:
@bshaffer Can you take a look?

php-nginx/build-scripts/install_php70.sh
```diff
@@ -0,0 +1,35 @@
...
+        gcp-php70-mailparse \
+        gcp-php70-memcached \
+        gcp-php70-mongodb \
+        gcp-php70-redis
+
+# Enable some extensions for backward compatibility
+${PHP70_DIR}/bin/php70-enmod apcu-bc
+${PHP70_DIR}/bin/php70-enmod apcu
+${PHP70_DIR}/bin/php70-enmod mailparse
+${PHP70_DIR}/bin/php70-enmod memcached
```
bshaffer: Shouldn't we be adding `ln -s ${PHP70_DIR} ${PHP_DIR}` here?

php-nginx/entrypoint.sh
```diff
@@ -133,9 +132,7 @@ chmod -R 550 ${DOCUMENT_ROOT}
 # Change the www-data's shell back to /usr/sbin/nologin
 chsh -s /usr/sbin/nologin www-data
 
-# Put a stricter php-cli.ini
-if [ -f "${PHP_DIR}/lib/php-cli-strict.ini" ]; then
-    mv ${PHP_DIR}/lib/php-cli-strict.ini ${PHP_DIR}/lib/php-cli.ini
-fi
+# Enable suhosin
+${PHP56_DIR}/bin/php56-enmod suhosin
```
bshaffer: What happens if `composer.json` requires `PHP 7.0`?

testapps/php56/tests/PHP56Test.php
```diff
@@ -53,13 +53,12 @@ public function testNonExistentPath()
...
     {
-        // Access to phpinfo.php, while phpinfo() is disabled by default.
+        // Access to phpinfo.php, while phpinfo() should be enabled by default.
         $resp = $this->client->get('phpinfo.php');
         $this->assertEquals('200', $resp->getStatusCode(),
                             'phpinfo.php status code');
-        $this->assertEquals('', $resp->getBody()->getContents(),
-                            'phpinfo() should be disabled and the content'
-                            . ' should be empty.');
+        $this->assertTrue(strlen($resp->getBody()->getContents()) > 1000,
```
bshaffer: Why is `phpinfo` enabled by default when suhosin is active? It's included in the blacklist above.

php-nginx/entrypoint.sh
tmatsuo: That's a good point, but this should be addressed in other PR #187

testapps/php56/tests/PHP56Test.php
tmatsuo: Because now we have `suhosin.simulation=On`

php-nginx/whitelist_functions.php
```diff
@@ -14,22 +14,33 @@
...
 }
 
-$suhosinBlacklist = ini_get('suhosin.executor.func.blacklist');
 $disable_functions = ini_get('disable_functions');
 
+// It's tricky, but we still don't have suhosin enabled and we can not fetch
+// the list from our suhosin config. Now it defines the list as a literal.
+// Note: We need to sync this list with our default suhosin config
+// (deb-package-builder/extensions/suhosin/debian/ext-suhosin.ini).
+// TODO: Find a better way
```
bshaffer: I am confused as to why we cannot pull this using `ini_get`. We are calling this in `entrypoint.sh`, which is well after the extensions are activated.

php-nginx/build-scripts/install_php70.sh
tmatsuo: No

testapps/php56/tests/PHP56Test.php
bshaffer: Oh, sorry, I got `testapp/php56` confused with `testapp/php56_custom`.

php-nginx/whitelist_functions.php
tmatsuo: That's not correct. Now we're activating suhosin just before running supervisord.

php-nginx/build-scripts/install_php70.sh
tmatsuo: We shouldn't do that because of the current flow.

Now we're installing PHP 5.6 and PHP 7.0, but still the default should be PHP 5.6.

bshaffer: Looking through the code, I understand it now. Essentially the container builds with php5.6 enabled, and then `composer.sh` gets executed to install the dependencies and change the php version if applicable.

Personally, I'd like to see a new script for enabling a PHP, i.e. `enable_php.sh`. And then remove `ln -s` from `install_php56.sh`. I understand right now we use `composer.json` alone to determine the version, but this might change. It could be split into `enable_php.sh` and `install_app_dependencies.sh`

php-nginx/whitelist_functions.php
bshaffer: Ok, I understand this now. In that case, can't we just call this after we enable the suhosin extension?

php-nginx/build-scripts/install_php70.sh
bshaffer: But this is a nit, I can open a bug for it :D

php-nginx/whitelist_functions.php
tmatsuo: Yeah, you're sane, but unfortunately no because

* If we enable suhosin in earlier stage, `composer run-script` fails because of suhosin. Suhosin is under simulation mode, but it dies because of `phar` is not allowed. Yeah, seems like it's a suhosin's bug, but now it works like that.

* and whitelist_functions.sh must be executed before `composer run-script` for some cases. An example is in php56_custom test app. The Sympony/Process needs `proc_open` and `proc_close` to be enabled.

bshaffer: Why can't we do (in `entrypoint.sh`):

1. Enable Suhosin
2. run `whitelist_functions.php`
3. run `composer run-script`

I think the reason is if suhosin is already enabled, `whitelist_functions.ini` won't be loaded? But in that case we can re-enable suhosin, which is not great, but probably better than hardcoding the ini values. Thoughts?

tmatsuo: It's not easily possible because composer fails when suhosin is enabled. We can add some flags to workaround that (like we did long tim ago)though. Also I think it's safer to run the script without enabling suhosin.

tmatsuo: When I looked at the old `composer.sh`, I think we should add some flags anyways. I'll update it in the next commit.

tmatsuo: Or just keep having php-cli.ini as we did

tmatsuo: Ok, I added back the `php-cli.ini` which is specific to the cli interface, so now we can simply run `whitelist_functions.sh` just after enabling suhosin. Thanks for the feedback.

tmatsuo:
@bshaffer Now it has a way simpler flow. PTAL

PR #187, opened 2017-01-11
Adding PHP 7.1

### LINKED ISSUE #130: [RFC] PHP 7.1

tmatsuo:
fixes #130


php-nginx/entrypoint.sh
```diff
@@ -133,9 +132,7 @@ chmod -R 550 ${DOCUMENT_ROOT}
 # Change the www-data's shell back to /usr/sbin/nologin
 chsh -s /usr/sbin/nologin www-data
 
-# Put a stricter php-cli.ini
-if [ -f "${PHP_DIR}/lib/php-cli-strict.ini" ]; then
-    mv ${PHP_DIR}/lib/php-cli-strict.ini ${PHP_DIR}/lib/php-cli.ini
-fi
+# Enable suhosin
+${PHP56_DIR}/bin/php56-enmod suhosin
```
bshaffer: so we enable suhosin when the container starts (i.e. in the entrypoint), which is fine with me, but I don't like how this is run even if the PHP version isn't php56. I guess the reasoning here is we'd have to run `detect_php_version` again, and as this doesn't harm anything we just leave it. Is that accurate?

tmatsuo: Yeah I understand. It will be addressed with #187

tmatsuo:
Rebased on top of #184 

Now `install_phpxx.sh` also copies the config files, makes it a default PHP.

bshaffer:
FWIW I'm still not a huge fan of how we install PHP7.x over top of PHP5.6 - seems like we should just install one version.

tmatsuo:
@bshaffer 

It installs PHP 7.x and removes PHP 5.6 after that, so there is only one version at the same time, although there is a short time window where it has 2 versions...

tmatsuo:
@cedricziel @bshaffer 

FYI, this image is now on production

PR #188, opened 2017-01-13
Set appropriate HTTPS envvar

### LINKED ISSUE #37: Nginx doesn't detect the upstream https connection

tmatsuo:
fixes #37

### DIFF
--- a/testapps/php56_custom_configs/nginx-custom.conf
+++ b/testapps/php56_custom_configs/nginx-custom.conf
@@ -37,14 +37,19 @@ http {
 
     keepalive_timeout  65;
 
+    map $http_x_forwarded_proto $fastcgi_https {
+        default '';
+        https on;
+    }
+
     upstream php-fpm {
         server 127.0.0.1:9000 max_fails=3 fail_timeout=3s;
     }
 
     server {
         listen       8080;
-        root /app;
-        index index.html index.htm index.php;
+        root /app
+        index index.php index.html index.htm;
 
         # Media: images, icons, video, audio, HTC
         location ~* \.(?:jpg|jpeg|gif|png|ico|cur|gz|svg|svgz|mp4|ogg|ogv|webm|htc)$ {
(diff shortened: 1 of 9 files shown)


testapps/php56_custom_configs/nginx-custom.conf
```diff
@@ -76,11 +81,28 @@ http {
...
 
-	location /hello {
-          return 200 'hello from nginx-custom.conf';
+        # Drop accesses for some files
+        location = /favicon.ico { access_log off; log_not_found off; }
+        location ~ /\.          { access_log off; log_not_found off; deny all; }
+        location ~ ~$           { access_log off; log_not_found off; deny all; }
+
+        location /hello {
+            return 200 'hello from nginx-custom.conf';
```
bshaffer: Seems like we should remove this?

testapps/php56_nginx_conf/nginx.conf
```diff
@@ -76,11 +81,28 @@ http {
...
         include /etc/nginx/conf.d/*.conf;
 
-	location /hello {
-          return 200 'hello from nginx.conf';
+        # Drop accesses for some files
+        location = /favicon.ico { access_log off; log_not_found off; }
+        location ~ /\.          { access_log off; log_not_found off; deny all; }
+        location ~ ~$           { access_log off; log_not_found off; deny all; }
+
+        location /hello {
```
bshaffer: This too. These are examples that shouldn't be shipped!

testapps/php56_custom_configs/nginx-custom.conf
tmatsuo: This is a config file for the test app. The test will access this path to make sure it is working.

testapps/php56_nginx_conf/nginx.conf
tmatsuo: This is a config for the test app.

tmatsuo:
FYI it's in prod

PR #190, opened 2017-01-17
Mention that the DOCUMENT_ROOT is changed to read-only

tmatsuo:
Superseding #29

@bshaffer PTAL

FYI @cedricziel


tmatsuo:
@chingor13 Can you take a look?

PR #193, opened 2017-01-19
Enable System V IPC family.

### LINKED ISSUE #191: Enable System V IPC family
with adding the following configure options
```
                --enable-sysvsem \
                --enable-sysvshm \
                --enable-sysvmsg \
```

tmatsuo:
fixes #191 

See also:
http://php.net/manual/en/book.sem.php


tmatsuo:
@chingor13 

Yeah, after I submit this, the process goes as follows:

1. Uploads the debian packages to our cloud-apt repo, as an `unstable` tag (so it's `gcp-php-runtime-jessie-unstable` or something like `gcp-php-runtime-jessie-20170120-1`.
2. Then #194 allows us to run our test suite with that specific repo on our jenkins server
3. If the above build succeeds, we'll be confident that the new deb packages are running OK
4. Re-tag the unstable repo as a stable
5. Run the jenkins job again for the actual release
6. Now we can add tests for checking if there's System V IPC functions are there

PR #194, opened 2017-01-20
Allow RUNTIME_DISTRIBUTION override

tmatsuo:
so that we can build with arbitrary apt repo.


tmatsuo:
@chingor13 

Yeah the process is described at: https://github.com/GoogleCloudPlatform/php-docker/pull/193#issuecomment-273962370

I'll run the build on our jenkins too, with the custom RUNTIME_DISTRIBUTION

tmatsuo:
Alright, jenkins build is green, and the generated image `gcr.io/google_appengine/php:2017-01-20_03_09` has the right apt configuration.

PR #195, opened 2017-01-20
Changed the name of the PHP debian packages.

tmatsuo:
Before this PR, the name of the PHP debian packages are like:

gcp-php-5.6.30
gcp-php-7.0.15

and they provide virtual packages like gcp-php56, gcp-php70.

It turned out it is problematic when we have multiple patch versions
in our repo.

For example, if we have these two packages:

gcp-php-7.0.14_7.0.14-2
gcp-php-7.0.15_7.0.15-1

in a single repo, the following command fails:

```
$ apt-get -y install gcp-php70
```

because when there are multiple candidates for a virtual package, apt
can't decide which to pick, so it fails.

So we need to change the package names as follows:

gcp-php56_5.6.30-1
gcp-php70_7.0.15-1

so that our repo will have only one package for each minor versions.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -11,6 +11,7 @@ machine:
     PHP_DOCKER_GOOGLE_CREDENTIALS: ${HOME}/credentials.json
     GOOGLE_PROJECT_ID: php-mvm-a
     TAG: circle-${CIRCLE_BUILD_NUM}
+    RUNTIME_DISTRIBUTION: gcp-php-runtime-jessie-unstable
 
 dependencies:
   override:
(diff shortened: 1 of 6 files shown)


tmatsuo:
The apt repo created with the generated debian packages as `gcp-php-runtime-jessie-20170120-3` and running our test with that repo.

tmatsuo:
I just realized that we can always run the test on CircleCI with our `gcp-php-runtime-jessie-unstable` repo so that we don't have to manually run our tests on our jenkins. I changed the CircleCI configuration and let's see how it goes.

tmatsuo:
@chingor13 Can you take a look?

I built the debian packages, upload them as `gcp-php-runtime-jessie-unstable`, and now the CircleCI tests are always running with that repo and its' green :)

PR #198, opened 2017-02-01
Add packages.yml configuration for building the php packages

chingor13:
Allows us the maintain the list of versions in code and utilize the same package build scripts as the ruby package builders.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 **/vendor
 .php_cs.cache
+deb-package-builder/pkg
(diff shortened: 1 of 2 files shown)


chingor13:
At the very least, I think it's important/useful to keep track of which true versions we're building/maintaining at a given time.

tmatsuo:
@chingor I agree with you in general, but I still prefer keeping the file internally.

It is a configuration file for an internal tool. Also the version numbers in this file don't always match
 the version numbers in our latest image.

PR #203, opened 2017-02-03
Runtime builder implementation

tmatsuo:
We're going to change how to generate `Dockerfile` for App Engine Flexible applications. Now gcloud sdk is running some python code to generate `Dockerfile` locally. This builder implementation is to replace this functionality with a cloudbuild step.

The `create-dockerfile` docker image just creates `Dockerfile` and `.dockerignore` if not present. The generated `Dockerfile` must have an environment variable based on the value of `runtime_config` -> `document_root` of users' `app.yaml`.


chingor13:
What's the overall goal this is trying to solve? It looks like it's trying to create a cloud build step for autogenerating a Dockerfile/.dockerignore if the files aren't present.

tmatsuo:
@chingor13 Alright, updated the description

builder/create-dockerfile/src/create_dockerfile.php
```diff
@@ -0,0 +1,82 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+require_once __DIR__ . '/vendor/autoload.php';
+
+use Symfony\Component\Yaml\Yaml;
+
+class GenFiles
```
chingor13: We could probably have unit tests for this class, but the functionality is pretty simple. If for whatever reason, we want to add more feature detection (extensions perhaps) from the `app.yaml`, we'd probably not want to need end-to-end tests for each feature.

tmatsuo: Added unit tests

tmatsuo:
@chingor13 Just added unit tests. PTAL

tmatsuo:
The failure
```
1) Google\Cloud\tests\EndToEndTest::testExec
GuzzleHttp\Exception\ServerException: Server error: `GET https://circle-306-dot-php-mvm-a.appspot.com/exec.php` resulted in a `503 Service Unavailable` response:
<!DOCTYPE html>
<html lang=en>
  <meta charset=utf-8>
  <meta name=viewport content="initial-scale=1, minimum-scale=1, w (truncated...)
```

The failure seems to be an issue within App Engine frontend (again) since there are no logs on the server.

tmatsuo:
Tests keep faling, sorry, but good thing is the test caches all the mistake I made ;P
I think this time it will pass

PR #207, opened 2017-02-06
Add Imagick extension

### LINKED ISSUE #197: Provide imagick extension
fixes https://github.com/GoogleCloudPlatform/php-docs-samples/issues/313

chingor13:
Fixes #197


tmatsuo:
The error log says:
```
Step #5: + phpunit
Step #5: This version of PHPUnit is supported on PHP 7.0 and PHP 7.1.
Step #5: You are using PHP 5.6.30 (/usr/bin/php).
```

Maybe you can rebase to the master then it will run

tmatsuo:
The code looks good. I think you'll need to add `extension=imagick.so` in `php.ini` file in the testapps.

chingor13:
Is it better to add it to the php.ini file or run the `php56-enmod` command?

tmatsuo:
For now add it to the `php.ini`. We'll add a convenient logic in one of our runtime builders with the `-enmod` command.

PR #212, opened 2017-02-10
Simple logging

### LINKED ISSUE #98: Integrate with the vm image log rotation instead of cron inside container
The system image should be rotating logs in /var/log/app_engine. We should integrate with that, and make any file path changes necessary for that to work.

tmatsuo:
Fixes #98 

Redirect the stdout/stderr logs of child processes to supervisord, then logs will be collected by the sidecar.
The access logs of nginx will be available on `stdout` section of the cloud logging, the errors will be on `stderr` section.

It doesn't use /var/log/app_engine directory any more. Now we can remove cron and logrotate because we don't need them now.


JustinBeckwith:
Much cleaner!  LGTM.

PR #217, opened 2017-02-15
Adding tests for checking if the current versions are the latest

tmatsuo:
Now it only checks micro versoins of php56, php70, php71, and nginx.
My plan is to run this on Travis with its Cron feature.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -38,4 +38,4 @@ install:
 - scripts/install_test_dependencies.sh
 
 script:
-- scripts/run_test_suite.sh
+- scripts/check_versions.sh
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,8 +21,15 @@ branches:
   only:
     - master
 
+notifications:
+  email:
+    recipients:
+      - php-cloud-eng@google.com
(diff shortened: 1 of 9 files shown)


tmatsuo:
I haven't run it on travis, but it successfully failed (detected our nginx is old) locally.

.travis.yml
```diff
@@ -38,4 +45,4 @@ install:
 - scripts/install_test_dependencies.sh
 
 script:
-- scripts/run_test_suite.sh
+- scripts/check_versions.sh
```
chingor13: So travis will no longer run our image tests?

tmatsuo: We used to use Travis, but migrated to CircleCI. We haven't been using Travis for a long time.

tmatsuo:
@chingor13 Thanks,I removed the `cloudbuild.yaml`, PTAL

.travis.yml
chingor13: Ok, so we'll start using it again just for the version checks? Looks like it's a new beta feature of Travis that Circle doesn't have.

tmatsuo: Ya

PR #219, opened 2017-02-16
Dumps the credentials before checking the existence

tmatsuo:


tmatsuo:
Travis was still failing. I temporary turned on Travis hook for PRs, so it will successfully fail.

tmatsuo:
I don't understand why, but travis doesn't pick up this PR.

PR #225, opened 2017-03-01
Stop enabling apcu extension because it's redundant

### LINKED ISSUE #224: PHP Warning:  Module 'apcu' already loaded in Unknown on line 0
Not sure where this is occurring, but on a `7.1` deployment of wordpress, I'm getting the following warning on every PHP command:

```
PHP Warning:  Module 'apcu' already loaded in Unknown on line 0
```

My php.ini looks like this:

```
extension=bcmath.so
zend_extension=opcache.so
```

As you can see, `apcu` is not being loaded, so I believe this error is happening in the base runtime.

tmatsuo:
fixes #224

### DIFF
--- a/scripts/install_test_dependencies.sh
+++ b/scripts/install_test_dependencies.sh
@@ -44,6 +44,10 @@ if [ -z "${GOOGLE_PROJECT_ID}" ]; then
     exit 1
 fi
 
+if [ -z "${CLOUDSDK_VERBOSITY}" ]; then
+    CLOUDSDK_VERBOSITY='none'
+fi
+
 # Install composer and defined dependencies
 which composer || \
     (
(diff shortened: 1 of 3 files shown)


tmatsuo:
@dpebot merge when green

scripts/install_test_dependencies.sh
```diff
@@ -44,6 +44,10 @@ if [ -z "${GOOGLE_PROJECT_ID}" ]; then
     exit 1
 fi
 
+if [ -z "${CLOUDSDK_VERBOSITY}" ]; then
+    CLOUDSDK_VERBOSITY='DEFAULT'
```
chingor13: Is this a valid option? https://cloud.google.com/sdk/gcloud/reference/

tmatsuo:
@chingor13 Good point. DEFAULT has been working for me for so long, but I'd use `none`

PR #226, opened 2017-03-01
Common integration test framework application

chingor13:
This test application conforms to the proposed common integration framework test interface. Currently, only the root and logging tests are implemented as the exceptions and monitoring are not yet implemented in the google/cloud php library. This test also is not currently enabled yet, due to credential escalation bugs in the test runner.


tmatsuo:
@dpebot merge when green

PR #229, opened 2017-03-06
Phalcon

chingor13:
This pull request does 2 things:

1. Allows the debian package builder to specify a comma-separated list of extensions to build as the second argument to the docker run command. This is useful for just building a new extension so you are not rebuilding every extension when you only need 1 or 2.
1. Adds the phalcon extension. I have written a sample application that can test the phalcon extension and it works, but we need to decide how to test the extensions going forward. If we test every extension for every version using end-to-end tests, they will take a very long time on every commit/PR.


tmatsuo:
Overall, looks nice.

I'm getting the following when I run it without the second argument:

```
Starting to download apcu_bc-1.0.3.tgz (7,245 bytes)
.....done: 7,245 bytes
File /build/apcu_bc-1.0.3.tgz downloaded
++ ls 'apcu_bc-beta-*.tgz'
++ sed 's/apcu_bc-beta-\([0-9][0-9]*\.[0-9][0-9]*\.[0-9][0-9]*\)\.tgz/\1/'
ls: cannot access apcu_bc-beta-*.tgz: No such file or directory
+ EXT_VERSION=
+ PACKAGE_VERSION=-7.1.2
+ PACKAGE_FULL_VERSION=-7.1.2-2
+ PACKAGE_DIR=gcp-php71-apcu-bc--7.1.2
+ mv apcu_bc-beta-.tgz gcp-php71-apcu-bc--7.1.2.orig.tar.gz
mv: cannot stat 'apcu_bc-beta-.tgz': No such file or directory
```

chingor13:
Ah, `pecl download apcu-beta` downloads the file as `apcu-[version]`.

chingor13:
@tmatsuo The issue with beta PECL packages is now fixed. PTAL when you get a chance.

jaitaiwan:
@chingor13 how does one make use of this?

chingor13:
Currently, it's not so simple. We've packaged it as a debian package in our cloud repository. You'll have to use a custom Dockerfile that will look something like:

```
FROM gcr.io/google-app-engine/php

RUN apt-get update && \
    apt-get install -y gcp-php70-phalcon && \
    ${PHP70_DIR}/bin/php70-enmod phalcon
```

If you're running 5.6, replace the `70` above with `56`. The official documentation for these are in progress.

We know this is nowhere close to ideal so we're also investigating easier ways to install these extensions so stay tuned for those.

tmatsuo:
Does it make sense to pre-install all of our extension debian packages until we have an easier solution?

chingor13:
We could probably do that. You would still need to run the `enmod` command or have a custom `php.ini` script to actually enable it.

chingor13:
See #237

PR #231, opened 2017-03-16
Install the apcu-bc's config file back

tmatsuo:
Not elegant, but it should work


tmatsuo:
The circleci is already running the tests against the unstable repo, and it failed just as expected.

I just uploaded new debian packages to the unstable repo and re-run the tests.

chingor13:
LGTM

PR #234, opened 2017-03-21
Add cassandra

chingor13:
This PR adds support for building the cassandra php extension. We have additional helpers for downloading libuv1(-dev) from debian jessie backports and compiling the cassandra-cpp-driver(-dev/-dbg) packages from source. We will host the libuv1 and cassandra-cpp-driver packages on cloud apt as well.

The test only checks if the extension is loaded (for now) because installing cassandra on debian 8 is already a hassle.


deb-package-builder/extensions/cassandra-cpp-driver/build.sh
```diff
@@ -0,0 +1,24 @@
+#!/bin/bash
+
+set -ex
+
+source ${DEB_BUILDER_DIR}/extensions/functions.sh
+
+echo "Building cassandra cpp driver"
+
+echo "deb http://ftp.debian.org/debian jessie-backports main" > /etc/apt/sources.list.d/jessie-backports.list
```
tmatsuo: Isn't it safe to do this in the `Dockerfile`? It may affect other builds depending the execution order.

deb-package-builder/extensions/cassandra/build.sh
```diff
@@ -0,0 +1,20 @@
...
+
+echo "Building cassandra for gcp-php${SHORT_VERSION}"
+
+PNAME="gcp-php${SHORT_VERSION}-cassandra"
+
+ls -t ${ARTIFACT_DIR}/libuv1_* | head -n 1 | xargs dpkg -i
+ls -t ${ARTIFACT_DIR}/libuv1-dev_* | head -n 1 | xargs dpkg -i
+ls -t ${ARTIFACT_DIR}/cassandra-cpp-driver_* | head -n 1 | xargs dpkg -i
+ls -t ${ARTIFACT_DIR}/cassandra-cpp-driver-dev_* | head -n 1 | xargs dpkg -i
+apt-get install -y libgmp-dev
```
tmatsuo: Can it be in the Dockerfile too?

deb-package-builder/extensions/cassandra-cpp-driver/build.sh
chingor13: I probably should just download the files from the backports repo rather than adding a source which could provide other packages that won't be available at runtime.

PR #235, opened 2017-03-21
Adding freetype support to the GD extension

### LINKED ISSUE #233: FreeType support for gd extension
I read the php-nginx README.md and it says

> `These extensions are compiled as shared (you need to enable with php.ini):`
> `GD (gd; with PNG, JPEG and FreeType support)`

but when i check my app's `php -i`, I see that FreeType for gd extension is not enabled.

Do I have to add anything other than
`extension=gd.so`
in my php.ini to enable the FreeType support?

tmatsuo:
fixes #233


tmatsuo:
The debian packages are already on unstable

PR #237, opened 2017-03-22
Automatically install all of our pre-built extensions

chingor13:
Automatically install our extensions built as debian packages when we install php.

This will allow the users to only have to enable them by adding a custom `php.ini` file rather than forcing them to create a Dockerfile.

### DIFF
--- a/php-nginx/build-scripts/install_php56.sh
+++ b/php-nginx/build-scripts/install_php56.sh
@@ -21,6 +21,9 @@ set -xe
 apt-get install -y \
         gcp-php56 \
         gcp-php56-apcu \
+        gcp-php56-cassandra \
+        gcp-php56-ev \
+        gcp-php56-event \
         gcp-php56-grpc \
         gcp-php56-imagick \
         gcp-php56-json \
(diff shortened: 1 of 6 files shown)


jaitaiwan:
So just clarifying, this will allow the extensions to be installed with a custom ini?

chingor13:
When the new runtime is released, yes. I'll create a [new release](https://github.com/GoogleCloudPlatform/php-docker/releases) when it's available (hopefully today).

chingor13:
@jaitaiwan This is [released](https://github.com/GoogleCloudPlatform/php-docker/releases/tag/2017-03-22-19-13). Official docs are still being worked on.

jaitaiwan:
Cheers @chingor13. I just pulled the latest docker image to test. Are the extensions being placed in a non-std folder? I get this error message when I try it: `PHP Startup: Unable to load dynamic library '/opt/php71/lib/x86_64-linux-gnu/extensions/no-debug-non-zts-20160303/phalcon.so'`

chingor13:
The phalcon project only currently supports PHP 5.5, 5.6, and 7.0 - so it's only available for app engine flex on 5.6 and 7.0. You can specify your PHP version in your composer.json file.

PR #240, opened 2017-03-23
Installs nginx from debian

### LINKED ISSUE #238: Install nginx from debian repository
We're locked on a specified version and are building from source each build. If we use the stable nginx from debian, we can get security patches and notifications from our package security scanner. This should also speed up base image builds slightly.

chingor13:
The debian stable version is `1.6.2-5+deb8u4` vs. `1.10.3` when compiling from source. Using the debian stable version allows us to leverage the security scanner for our docker runtimes and have quicker security patches. The base image build should be faster as well (no compiling and no longer need to install/uninstall dev packages).

Additionally, we no longer need the gpg checks on when downloading the source, so they have been removed.

One potentially breaking change is that the nginx directory has changed. If a user has hard-coded paths to the nginx binary or to the nginx config directory, they will no longer work. Would could symlink the old locations to the new ones, but those paths should not have been considered stable to begin with.

Fixes #238


check-versions/tests/VersionTest.php
```diff
@@ -108,7 +108,7 @@ public function testNginxVersion()
     {
```
tmatsuo: Oh, can you remove `testNginxVersion` and other nginx related stuff in this file.

PR #242, opened 2017-04-06
Made the build_images.sh more re-usable

tmatsuo:


circle.yml
```diff
@@ -13,6 +13,7 @@ machine:
     TAG: circle-${CIRCLE_BUILD_NUM}
     RUNTIME_DISTRIBUTION: gcp-php-runtime-jessie-unstable
     BUILDER_TARGET_IMAGE: gcr.io/${GOOGLE_PROJECT_ID}/php-nginx
+    RUN_E2E_TESTS: true
```
chingor13: Will the jenkins build skip end to end tests then?

tmatsuo: First I added the env on the jenkins config page, but yeah it's better to have in the script, and it's done.

tmatsuo:
Hm. The e2e test is failing, but there's no any logs on the server side. Let's re-run the test.

PR #245, opened 2017-04-10
Install extensions

### LINKED ISSUE #25: PHP Extensions should be implicitly installable through composer.json
In order to maintain dependencies to shared PHP extensions, composer packages may define `ext-*` dependencies, like `ext-mbstring`-just like they may define a dependency on a interpreter-version.

The image build-process should be able to read those and add the corresponding `extension` directive to the top-level `php.ini` so that the developer doesn't need to add a `php.ini` file for already explicitly defined dependencies.

chingor13:
Adds a php script to enable extensions that have already been installed but not enabled. We add the extensions directives into `/opt/php/lib/conf.d/extensions.ini` (a folder where php looks for additional ini files).

We look in the composer.json for requires "ext-[extension name]" and enable them if they are available. This runs after we've installed the correct php version but before we try and run `composer install` (it will fail for missing extensions).

Also adds some unit tests for detecting php versions and detecting extensions to install.

Fixes #25


php-nginx/detect_php_version.php
```diff
@@ -18,32 +18,49 @@
...
+        }
+        return '';
+    }
 
-if (count($argv) < 2) {
-    die("Usage:\n" . $argv[0] . " filename\n");
+    private static function detectAvailableVersions()
+    {
+        return [
+            getenv('PHP71_VERSION'),
```
tmatsuo: Note: These envvars might be out of sync with the debian packages.

chingor13: They already are out of date, this is just ported to a class to make testing it easier.

php-nginx/install_extensions.php
```diff
@@ -0,0 +1,133 @@
...
+            getenv('PHP_DIR'),
+            'lib',
+            'conf.d',
+            'extensions.ini'
+        ], '/');
+    }
+
+    private function addExtension($package, $version)
+    {
+        if (in_array($package, self::SHARED_EXTENSIONS)) {
```
chingor13: This separate handling will be removed in favor of treating all the extensions the same way.

testapps/php56_extensions/composer.json
```diff
@@ -1,6 +1,33 @@
 {
     "require": {
         "php": "^5.6",
-        "phpunit/phpunit": "4.8.*"
+        "phpunit/phpunit": "4.8.*",
```
tmatsuo: What happens if you specify older version of an extension? My guess is that it will fail meeting dependencies when running composer. Does it make sense if we fail faster than that, ideally with an actionable error message?

chingor13: You mean if the developer puts a required version for the extension? Every example I've found of people using this field use "*" for the required version because extension versioning is inconsistent across extensions.

chingor13:
If we don't have the requested extension installed, the install script silently fails. The next step in the build is to run `composer install` which fails with a message like:

```
Loading composer repositories with package information
Updating dependencies
Your requirements could not be resolved to an installable set of packages.

  Problem 1
    - The requested PHP extension ext-phalcon * is missing from your system. Install or enable PHP's phalcon extension.
```

testapps/php56_extensions/composer.json
tmatsuo: For example, if you put

```
"ext-memcached": "1.0.2"
```

The install script will enable the memcache extension, but I think the `composer install` will fail later. The build log says "The requested PHP extension ext-memcached 1.0.2 is missing from your system."

Does it make sense for us to rather say "We always use the latest version of extensions, please use "*" for your extension dependencies" or something?

chingor13:
If a user already had a custom php.ini file and ext requires in the composer.json, they will see additional warnings in the logs about already loading extension-name.so.

tmatsuo:
@chingor13 

> If a user already had a custom php.ini file and ext requires in the composer.json, they will see additional warnings in the logs about already loading extension-name.so.

Is it possible for the script to detect if it's already loaded, and if it's loaded, it can stop processing the extension in question?

chingor13:
@tmatsuo Not at this time. The user's php.ini config isn't put into place until `entrypoint.sh`. When we move this config to build-time, we could check if it's already configured.

tmatsuo:
@chingor13 

> @tmatsuo Not at this time. The user's php.ini config isn't put into place until entrypoint.sh. When we move this config to build-time, we could check if it's already configured.

Sounds good

testapps/php56_extensions/composer.json
chingor13: It's not entirely accurate that we always have the latest versions of extensions. I think it would make sense to bail out early and tell them to specify "*" as the version because we don't support arbitrary versions.

tmatsuo: True!

php-nginx/install_extensions.php
```diff
@@ -0,0 +1,148 @@
...
+        'xmlrpc',
+        'xsl',
+        ## Debian package extensions below
+        'cassandra',
+        'ev',
+        'event',
+        'grpc',
+        'imagick',
+        'mailparse',
+        'memcache',
```
tmatsuo: FYI, we don't provide `memcache` for PHP 7.0 and 7.1

chingor13: Yep, I'll add a blacklist instead of maintaining a huge whitelist. Phalcon also isn't supported on 7.1 yet.

chingor13:
@tmatsuo Fixed the supported extension list. PTAL when you get a chance.

php-nginx/build-scripts/install_php56.sh
```diff
@@ -40,6 +40,8 @@ apt-get install -y \
         --no-install-recommends
 
 # Enable some extensions for backward compatibility
+ln -s ${PHP56_DIR}/bin/php56-enmod ${PHP56_DIR}/bin/php-enmod
```
tmatsuo: No action required: If we don't use these command any more, maybe we can eventually stop creating these links, also remove them from the debian packages.

php-nginx/README.md
```diff
@@ -224,18 +224,27 @@ php.ini):
...
 - GD (gd; with PNG, JPEG and FreeType support)
 - gettext (gettext)
-- gRPC (shared)
+- gRPC (grpc)
 - intl (intl)
+- ImageMagick (imagick)
 - mbstring (mbstring)
-- memcached (shared, enabled by default, experimental support for PHP7)
-- mongodb (shared)
+- memcached (enabled by default, experimental support for PHP7)
```
tmatsuo: I think we can remove `experimental support for PHP7` because we started using a stable versionof memcached from pecl

PR #247, opened 2017-04-13
Added sleep after the deployment for e2e tests.

tmatsuo:


testapps/php71_e2e/tests/tests/EndToEndTest.php
```diff
@@ -69,6 +69,8 @@ public static function setUpBeforeClass()
             );
         }
         self::deploy($project_id, $e2e_test_version);
+        echo 'Sleeping for the deployment settles down' . PHP_EOL;
+        sleep(60);
```
chingor13: This seems a little heavy handed. Can we try and hit the health check (or a different page) a few times (with retry) to see if the app is up? Or is there something else that we're waiting for?

tmatsuo: I see some cases where there is a failure after few successful requests (502), so currently it's not enough to see a single success.

I'll try to introduce retries on the tests themselves.

PR #252, opened 2017-04-13
Fix image name for jenkins build.

chingor13:
We renamed the test image to php from php-nginx and updated the circle ci config previously.

This fixes the jenkins build (for testing new package versions)


chingor13:
Wait, we shouldn't even need to retag the image to the production namespace.

PR #254, opened 2017-04-17
Get the yaml path from the envvar and add the pipeline definition.

tmatsuo:


builder/create-dockerfile/src/GenFiles.php
```diff
@@ -23,21 +23,28 @@ class GenFiles
...
+    private $appYamlPath;
+
     /**
      * Constructor allows injecting the workspace directory.
      */
     public function __construct($workspace = self::DEFAULT_WORKSPACE)
     {
         $this->workspace = $workspace;
+        $yamlPath = getenv('GAE_APPLICATION_YAML_PATH') !== false
+            ?: self::DEFAULT_YAML_PATH;
```
chingor13: won't `$yamlPath` be true if the env variable is set?

tmatsuo: Yes, you're right. I added a test and it should fail.

tmatsuo: Failed as expected
```
RuntimeException: The application yaml file does not exist: /tmp/GenFilesTesteLdIFk/1
```

BTW true becomes `1`!

This time the test should pass.

tmatsuo:
@chingor13 Thanks! Now it's passing. PTAL

PR #258, opened 2017-04-20
Have pcntl compiled in, disabled opcache for cli by default

tmatsuo:


tmatsuo:
The test should fail now.

tmatsuo:
Packages uploaded to unstable repo, so restarted the test

PR #270, opened 2017-05-04
Use envvar DOCUMENT_ROOT in nginx.conf

tmatsuo:
This is an attempt to stop replacing the document root in `entrypoint.sh`.


tmatsuo:
The only concern is the nginx binary in `nginx-extras` is little bit bigger than the one in `nginx-full` and it will use a little bit more memory.

tmatsuo:
I did a quick performance test (it's still manual), and there's no sign of significant performance degradation.

tmatsuo:
Yeah, it's possible and it's a breaking change! We should state it on our release note.

tmatsuo:
It's a BC break and I notified potential users.
https://github.com/viender/viender/issues/3
https://github.com/a1comms/cf-gae-flex-php/issues/1

PR #273, opened 2017-05-09
Build runtime images and run tests in parallel

chingor13:
By using a Container Builder pipeline, we can reduce our test time from ~37 minutes to ~24 minutes (on CircleCI. The e2e test takes ~12 minutes by itself).

The main cloudbuild.yaml now builds the test runner, php-base image, php-onbuild image, and the gen-dockerfile builder.

Each of these images' tests are also run as soon as their respective image is built.


chingor13:
Parallelizing the build uncovered an race condition using the structure_test image in parallel. https://github.com/GoogleCloudPlatform/runtimes-common/pull/169

cloudbuild.yaml
```diff
@@ -0,0 +1,194 @@
...
+    id: php71-extensions-build
+  - name: gcr.io/cloud-builders/docker
+    args: ['run', 'gcr.io/$PROJECT_ID/php71-extensions:$_TAG', 'vendor/bin/phpunit']
+    dir: testapps/php71_extensions
+    waitFor: ['php71-extensions-build']
+
+images:
+  - gcr.io/$PROJECT_ID/php-base:$_TAG
+  - gcr.io/$PROJECT_ID/php:$_TAG
+  - gcr.io/$PROJECT_ID/php-base:$_TAG
```
tmatsuo: gen-dockerfile?

chingor13: will fix

builder/gen-dockerfile/src/create_dockerfile.php
```diff
@@ -18,8 +18,10 @@
 require_once __DIR__ . '/../vendor/autoload.php';
 require_once __DIR__ . '/GenFiles.php';
 
-$genFiles = new GenFiles();
-if ($argc === 2) {
+$genFiles = $argc >= 3
```
chingor13: We can use option parsing in a separate PR

tmatsuo: I have a change with Symfony\Console

chingor13:
We can also run the end to end test in parallel with other tests if it always runs. It's currently possible to skip the e2e tests, but both circle and our release builder are configured to run them.

chingor13:
@tmatsuo can you take a look at the last few changes since the merge from master. Switched to using the Symfony Console CommandTester.

tmatsuo:
LGTM

tmatsuo:
Oh wait. Do we need to change `builder/php-latest.yaml`?

chingor13:
I left the base image as the first argument and made the workspace an option.

PR #274, opened 2017-05-09
Let the builder add env vars.

tmatsuo:
* Our BUILDER runs the composer script
* env_variables in app.yaml will be in the final Dockerfile
* some values in runtime_config will be transformed to an envvar
* values in runtime_config always win


php-base/composer.sh
```diff
@@ -77,10 +77,14 @@ EOF
     php -d auto_prepend_file='' /tmp/install_extensions.php ${APP_DIR}/composer.json ${PHP_DIR}/lib/conf.d/extensions.ini ${PHP_VERSION}
 
     # Run Composer.
+    if [ -n "${GOOGLE_RUNTIME_RUN_COMPOSER_SCRIPT}" ]; then
```
tmatsuo: Maybe we'll remove this when we remove the `php` image.

`--no-script` introduced for a user who was suffered from a fact that our build pipeline doesn't respect env_variables in app.yaml. 

So yes, when the builder is used, it always run the composer script.

builder/gen-dockerfile/src/GenFilesCommand.php
```diff
@@ -0,0 +1,158 @@
...
+
+        $envs = $this->envsFromRuntimeConfig()
+            + $this->envsFromAppYaml()
+            + [
+                'DOCUMENT_ROOT' => self::APP_DIR,
+                'GOOGLE_RUNTIME_RUN_COMPOSER_SCRIPT' => 'true'
+            ];
+        $envString = 'ENV ';
+        foreach ($envs as $key => $value) {
+            $envString .= "$key=$value \\";
```
chingor13: Looks like this needs a newline

PR #277, opened 2017-05-11
Test runner migration

tmatsuo:


cloudbuild-test-runner/Dockerfile
```diff
@@ -14,32 +14,29 @@
 
 # Dockerfile for running phpunit within a cloudbuild step.
 
-FROM python:2.7-alpine
+FROM gcr.io/google-appengine/php-base:latest
+
+RUN /bin/bash /build-scripts/install_php56.sh
```
chingor13: Do we care about it being php 5.6?

chingor13: Not a blocker, we can probably use the default 7.1 in the future

tmatsuo: Now it fails for testapps/php56_nginx_conf/composer.json, saying "This package requires php ^5.6 but your PHP version (7.1.4) does not satisfy that requirement". Let's upgrade it in the future.

PR #278, opened 2017-05-11
Created versioned php images

chingor13:
* Add php71, php70, php56 images
* php-base no longer contains php
* php-onbuild is built from php71


php-versioned/Dockerfile
```diff
@@ -0,0 +1,28 @@
+# Copyright 2017 Google Inc.
```
tmatsuo: Unintentionally added?

chingor13: yup

PR #279, opened 2017-05-11
update php 7.1.5 and 7.0.19 for building debian packages

chingor13:
Our automated builder will build all the new packages next time it runs.


tmatsuo:
Oh I noticed there's grpc 1.3.2
https://pecl.php.net/package/gRPC/1.3.2

Does it make sense to add it too?

tmatsuo:
Oh never mind, it will likely be picked up

chingor13:
Yeah we aren't locking the grpc version

PR #281, opened 2017-05-12
Integration & Acceptance tests

chingor13:
Adds a new cloudbuild pipeline for the e2e/integration tests. We run the pipeline against the images we just built (including the builder pipeline) after building the images.

Added a new `scripts/run_acceptance_tests.sh` which runs the same integration tests but with the production images and the production build pipeline.


testapps/build_pipeline/app.yaml
```diff
@@ -0,0 +1,8 @@
+runtime: php
+env: flex
+
+runtime_config:
+  document_root: web
+
+env_variables:
+  WHITELIST_FUNCTIONS: phpinfo,exec
```
tmatsuo: Can it be in the runtime_config?

```
runtime_config:
  document_root: web
  whiltelist_functions: phpinfo,exec
```

testapps/build_pipeline/nginx-app.conf
```diff
@@ -0,0 +1,18 @@
+# Copyright 2015 Google Inc.
```
tmatsuo: I think we don't need this file

tmatsuo: I don't think we need this file

testapps/build_pipeline/app.yaml
chingor13: I think we should move it here once we switch to the builders. All the other examples use env variables for now.

testapps/build_pipeline/nginx-app.conf
chingor13: Ok, I'll remove it. This was copied from the e2e test.

testapps/build_pipeline/app.yaml
chingor13: Ah, I see. The runtime builder already implemented this, so we're making sure that we're using the builder.

PR #283, opened 2017-05-12
Fix the version checker test

chingor13:
cloudbuild-test-runner is now built from the php onbuild image and no longer contains docker.
run the apt commands from inside the cloudbuild-test-runner.


tmatsuo:
Thanks for catching this

Issue #262, opened 2017-04-21
Custom Config in "http" Block (nginx)


iamacarpet:
Hello,

Love the way you can override the nginx config using nginx-app.conf for stuff inside the "server" block, or nginx.conf to replace the whole thing.

I've found a need to add things inside the "http" block but outside the "server" block ([1](https://github.com/a1comms/cf-gae-flex-php/blob/master/nginx.conf)), so currently I'm having to include an nginx.conf and replicate/replace the whole thing.

Would you consider supporting an nginx-http.conf similar to nginx-app.conf, but included in a different place in the main config?

Hopefully that'd mean there was less chance of some other change breaking things than if I was fully over-writing the main nginx config.

Regards,
iamacarpet

tmatsuo:
@iamacarpet Yeah it does make sense. We'll add this feature. Just out of curiosity, what configuration you're adding?

iamacarpet:
@tmatsuo Thanks very much. I'm adding some usage of the realip module and the geo module, to work around the REMOTE_ADDR bug [1], plus 403 if the requests don't come via Cloudflare (copy of the full config here [2]).

[1]: https://issuetracker.google.com/issues/36120924
[2]: https://github.com/a1comms/cf-gae-flex-php/blob/master/nginx.conf

PR #285, opened 2017-05-13
front_controller_file support in runtime_config

tmatsuo:
Note to @chingor13 it's not urgent at all, just review next week


tmatsuo:
Probably we should add a support for #262 too with this PR

chingor13:
Do we know that the variable we set in the server section propagates to the `nginx-app.conf` file? I don't think we have a test that actually checks that the front controller file is being hit.

tmatsuo:
It's another BC break, I notified the current users.

https://github.com/viender/viender/issues/3
https://github.com/a1comms/cf-gae-flex-php/issues/1

PR #289, opened 2017-05-16
Allow acceptance test to use default pipeline

chingor13:
by allowing RUNTIME_BUILDER_ROOT env to be an empty string.

In PHP: 

```php
var_dump("" == false);
=> bool(true)

var_dump("" === false);
=> bool(false)
```


chingor13:
The other env variable checks can stay as `==` because they are not allowed to be empty either.

PR #291, opened 2017-05-17
Build pipeline base image

chingor13:
The php-base image no longer has php installed, so we need to specify a versioned (non-on-build) image until gen-dockerfile is updated to detect eh php version and pick the correct base image.

Also adds a test that we install the correct php version as we're using the php71 image in the builder pipeline.

Additionally, we will have to publish the php versioned images in gcr.io/google-appengine namespace.


tmatsuo:
LGTM

My only concern is since there are many deployments happening, so if we run our test cases simultaneously (like for PR and for push, or on circle and on our jenkins), it will fail. Let's have it as it is now, but eventually it's better to have tests only for the builder part.

chingor13:
Makes sense. Additionally, we can have leftover services running if a test fails as all the parallel processes are terminated if one fails (the app deletion never happens).

PR #294, opened 2017-05-19
Add a new envvar for the http block in nginx.conf

### LINKED ISSUE #262: Custom Config in "http" Block (nginx)

tmatsuo:
will fix #262

### DIFF
--- a/php-base/Dockerfile
+++ b/php-base/Dockerfile
@@ -90,9 +90,8 @@ RUN mkdir -p $APP_DIR $UPLOAD_DIR $SESSION_SAVE_PATH \
     && ln -sf ${PHP_DIR}/bin/php /usr/bin/php
 
 # Put other config and shell files into place.
-COPY nginx.conf fastcgi_params gzip_params $NGINX_DIR/
-COPY nginx-app.conf $NGINX_USER_CONF_DIR
-RUN touch "${NGINX_USER_CONF_DIR}/nginx-http.conf"
+COPY nginx.conf fastcgi_params gzip_params "${NGINX_DIR}/"
+COPY nginx-app.conf nginx-http.conf "${NGINX_USER_CONF_DIR}/"
 COPY supervisord.conf /etc/supervisor/supervisord.conf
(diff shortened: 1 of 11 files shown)


php-base/Dockerfile
```diff
@@ -92,6 +92,7 @@ RUN mkdir -p $APP_DIR $UPLOAD_DIR $SESSION_SAVE_PATH \
 # Put other config and shell files into place.
 COPY nginx.conf fastcgi_params gzip_params $NGINX_DIR/
 COPY nginx-app.conf $NGINX_USER_CONF_DIR
+RUN touch "${NGINX_USER_CONF_DIR}/nginx-http.conf"
```
chingor13: Not a blocker, but if we add an empty file and combine it with the above COPY, it's one fewer docker layer.

php-onbuild/README.md
```diff
@@ -294,7 +294,10 @@ various config files" below).
 There are environment variables for adding config files.
 
 - NGINX_CONF_INCLUDE (defaults to nginx-app.conf)
-  This file will be inserted to the main nginx configuration.
+  This file will be inserted at the server block in the main nginx
+  configuration.
+- NGINX_CONF_HTTP_INCLUDE (defaults to nginx-http.conf)
+  This file will be inserted at the http block in the main nginx configuration.
```
chingor13: Maybe note that this is only the case for our default `nginx.conf` file. If they provide their own, it won't work unless they add the include line as well.

Issue #239, opened 2017-03-23
Move config file overrides to build time.


chingor13:
Currently, we move the `nginx-app.conf`, `nginx.conf`, `php-fpm.conf`, `php.ini`, `additional-supervisord.conf`, and `supervisord.conf` files in the entrypoint of the runtime.

These environment variables/config values shouldn't change between build and run time.

We also overwrite the document root in the nginx.conf, but this also shouldn't change between build and run time.

garnold:
[Related](https://github.com/GoogleCloudPlatform/php-docker/issues/295)

tmatsuo:
It's fixed with the new builder pipeline.

tmatsuo:
the builder pipeline

PR #297, opened 2017-05-26
Removed the entrypoint.sh from the versioned images

### LINKED ISSUE #295: Support volume mounting for local development
Hello, digging GAE PHP FE thus far!

One challenge we're facing during local development is we'd like to be able to mount our application directory onto the container so we don't need to rebuild when we change files. Something like:

```
docker run --name <name> --publish 8080:8080 --rm --volume=`pwd`:/app <image>:latest
```

The issue is `entrypoint.sh` messes with file permissions and moves files around, causing unexpected changes to our local filesystem. 

[The docs recommend testing applications outside of Docker](https://cloud.google.com/appengine/docs/flexible/php/quickstart#run_hello_world_on_your_local_machine), however that defeats the purpose of using Docker in the first place IMHO.

The fix would be to move the majority of `entrypoint.sh` to `Dockerfile`, except for the Composer stuff. We're happy to rebuild when we change core configuration files like `php.ini` and `nginx*.conf`, but rebuilding when changing application files is tedious.

Thanks for your consideration!

tmatsuo:
It will fix #295


chingor13:
This will also fix #239 when we switch to runtime builders

Issue #303, opened 2017-06-01
ext-gmp Plans?


ConnorVG:
Are there any plans, currently, to add support for `php-gmp`?

tmatsuo:
We didn't have a plan, but now we have because of your request.

@chingor13 Can you take a look?

chingor13:
Yep will do. I'm not seeing an major concerns about this extension so maybe we'll just compile it into PHP (rather than as shared). WDYT?

tmatsuo:
SG

ConnorVG:
I think this would be awesome – having to have a `Dockerfile` that simply installs GMP is a bit lame. Shared is 100% fine, there is nothing special about needing it as an external extension (afaict).

chingor13:
The gmp extension should now be available by default in the runtime for new deployments as of [today's release](https://github.com/GoogleCloudPlatform/php-docker/releases/tag/2017-06-06-13-41).

ConnorVG:
Amazing.

PR #304, opened 2017-06-05
Add the gmp extension, included by default.

chingor13:
See #303


chingor13:
The latest packages are pushed to the unstable debian repository to allow testing to work.

PR #308, opened 2017-06-08
Added runtime_config to env var mapping

tmatsuo:


builder/gen-dockerfile/tests/test_data/values_on_both/app.yaml
```diff
@@ -0,0 +1,24 @@
...
+  front_controller_file: app.php
+  nginx_conf_http_include: files/nginx-http.conf
+  nginx_conf_include: files/nginx-app.conf
+  nginx_conf_override: files/nginx.conf
+  php_fpm_conf_override: files/php-fpm.conf
+  php_ini_override: files/php.ini
+  supervisord_conf_addition: files/additional-supervisord.conf
+  supervisord_conf_override: files/supervisord.conf
+
+env_variables:
```
chingor13: At runtime, these will be set by App Engine and overwrite the ones set in the runtime config. Is this something we want to warn or even fail the build for?

chingor13:
Also, maybe update the README.md?

builder/gen-dockerfile/tests/test_data/values_on_both/app.yaml
tmatsuo: Yeah let's fail when both have values

tmatsuo: Done. Now it fails when there are values on both.
@chingor13 PTAL

PR #310, opened 2017-06-09
Compile 7.1.6 by default

tmatsuo:


tmatsuo:
Hm, seems like I need to update the unstable apt repo for the test to pass...

tmatsuo:
Uploaded packages to "unstable" repo

PR #313, opened 2017-06-20
Build 7.2.0 alpha debian packages

chingor13:
Add support for building 7.2.0alphaX.

- [x] Cannot install PEAR (and thus cannot install PECL)
- [x] Ensure we can build our extensions on 7.2.0


chingor13:
Creates a php72 image and compiles the debian packages for 7.2.0alpha3.

We won't release the real php72 image until December when it's released.

PR #320, opened 2017-06-27
Added a script for collecting deployment latency of simple apps.

tmatsuo:
It doesn't need to be a phpunit test, but it's easy to utilize our test runner.

### DIFF
--- a/perf-dashboard/deployment-latency/tests/CollectDeploymentLatencyTest.php
+++ b/perf-dashboard/deployment-latency/tests/CollectDeploymentLatencyTest.php
@@ -21,7 +21,6 @@ use Google\Cloud\BigQuery\BigQueryClient;
 
 class CollectDeploymentLatencyTest extends \PHPUnit_Framework_TestCase
 {
-
     const DEPLOYMENT_MAX_RETRY = 5;
     const PROJECT_ID = 'php-perf-dash';
     const DATASET_ID = 'deployment_latency';
@@ -124,7 +123,7 @@ class CollectDeploymentLatencyTest extends \PHPUnit_Framework_TestCase
             ]
         );
         $url = 'http://metadata.google.internal/computeMetadata/'
-            .'v1/instance/service-accounts/default/token';
+            . 'v1/instance/service-accounts/default/token';
         $token = json_decode(file_get_contents($url, false, $context), true);
 
         return new BigQueryClient(
--- a/perf-dashboard/deployment-latency/tests/CollectDeploymentLatencyTest.php
+++ b/perf-dashboard/deployment-latency/tests/CollectDeploymentLatencyTest.php
@@ -1,0 +1,159 @@
+<?php
+/**
+ * Copyright 2017 Google Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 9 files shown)


perf-dashboard/deployment-latency/tests/CollectDeploymentLatencyTest.php
```diff
@@ -0,0 +1,158 @@
...
+                $latency = 0.0;
+                while ($failureCount < self::DEPLOYMENT_MAX_RETRY) {
+                    $ret = self::execWithResult($command);
+                    if ($ret === 0) {
+                        $latency = microtime(true) - $before;
+                        break;
+                    }
+                    $failureCount++;
+                }
+                self::$bigQuery = self::createBigQueryClient();
```
chingor13: not a huge deal, but this could move to the `setUpBeforeClass`

perf-dashboard/deployment-latency/tests/CollectDeploymentLatencyTest.php
```diff
@@ -0,0 +1,158 @@
...
+            'composer require php'
+        );
+        return $tempDir;
+    }
+
+    public static function setUpBeforeClass()
+    {
+        self::execWithError(
+            'gcloud config set app/use_runtime_builders true',
+            'use-runitme-builder'
```
chingor13: typo in the error message

perf-dashboard/deployment-latency/tests/CollectDeploymentLatencyTest.php
tmatsuo: Since we're fetching the access token directly as a workaround, the token will be expired for later bigquery calls if we do it in `setupBeforeClass`, although I haven't tested it.

perf-dashboard/deployment-latency/tests/CollectDeploymentLatencyTest.php
```diff
@@ -0,0 +1,158 @@
...
+                    $failureCount++;
+                }
+                self::$bigQuery = self::createBigQueryClient();
+                $dataset = self::$bigQuery->dataset(self::DATASET_ID);
+                $table = $dataset->table(self::TABLE_ID);
+                $timestamp = self::$bigQuery->timestamp(new \DateTime());
+                $row = [
+                    'deployment_latency_seconds' => $latency,
+                    'report_name' => $reportName,
+                    'failure_count' => $failureCount,
```
chingor13: Would it make more sense to maybe add a test id and report each attempt separately (with an attempt #)? Normal gcloud deploys don't retry so if you're querying for average deploy times you'd have to exclude ones where failure_count != 0

perf-dashboard/deployment-latency/tests/CollectDeploymentLatencyTest.php
tmatsuo: done

perf-dashboard/deployment-latency/tests/CollectDeploymentLatencyTest.php
tmatsuo: Good point. I made it to reset the timer upon failure.

PR #321, opened 2017-06-27
Increased the timeout of the Container Builder

tmatsuo:


scripts/record_deployment_latency.sh
```diff
@@ -36,6 +36,7 @@ if [ -n "${REBUILD_TEST_RUNNER}" ]; then
            cloudbuild-test-runner
 fi
 
-gcloud -q container builds submit  perf-dashboard/deployment-latency\
+gcloud -q container builds submit perf-dashboard/deployment-latency\
+       --timeout 7200 \
```
chingor13: That's a long time :)

PR #322, opened 2017-06-29
Disallow apps without document_root configured

tmatsuo:


chingor13:
Do we even want to allow the user to specify `/app` as the document root?

tmatsuo:
> Do we even want to allow the user to specify /app as the document root?

In some cases, it might be cumbersome to change the directory structure of your app. I'd provide a way to opt-in to serve the entire app for such cases.

PR #328, opened 2017-07-06
Removed the PHP micro versions in the structure test

tmatsuo:
I think we need to streamline the process of the full release...


tmatsuo:
Seems like circleCI has some troubles. Let's merge this and use our jenkins server for the release.

PR #329, opened 2017-07-11
[WIP] First trial for upgrading to debian 9 (Stretch)

tmatsuo:
Mainly for the record

Stretch has openssl 1.1 and some packages have an issue compiling against it.

Related issues:
- cassandra-cpp-driver: https://datastax-oss.atlassian.net/browse/CPP-438
- php 5.6: https://github.com/oerdnj/deb.sury.org/issues/566
- curl header: https://github.com/phpbrew/phpbrew/issues/861

### DIFF
--- a/deb-package-builder/Dockerfile
+++ b/deb-package-builder/Dockerfile
@@ -18,14 +18,18 @@
 # Then you'll get deb packages in /mydir.
 
 # Inherit the php base image so we use the same dependency libraries
-FROM gcr.io/google-appengine/php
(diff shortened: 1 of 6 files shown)


jaitaiwan:
@tmatsuo I've just done a little sticky beak and it looks like the top two issues you provided are resolved and the last one seems to have a work around and be fixed in [some version of php](https://bugs.php.net/bug.php?id=74125)

Hope that helps!

chingor13:
Closed in favor of #432

PR #338, opened 2017-07-21
Use the runtime builder for the integration test

tmatsuo:


tmatsuo:
@nkubala 

Good point, to test both, I think we need to have another test app, because for xrt, we need to use `runtime: custom`. However I'm more concerned with testing the runtime builder because xrt will go away soon.

Anyways, even if we decide to test xrt, I'd do that in another PR. WDYT?

PR #343, opened 2017-08-03
Adding TEST_VM_IMAGE flag for testing against different VM images

tmatsuo:


tmatsuo:
I don't understand the logs for the test failure, so I just restarted the test.

integration-tests.yaml
```diff
@@ -26,6 +26,7 @@ steps:
     - 'TAG=${_TAG}-build-pipeline'
     - 'E2E_PROJECT_ID=${_E2E_PROJECT_ID}'
     - 'RUNTIME_BUILDER_ROOT=${_RUNTIME_BUILDER_ROOT}'
+    - 'TEST_VM_IMAGE=${_TEST_VM_IMAGE}'
```
chingor13: `build_images.sh` also uses this cloudbuild yaml so that `gcloud container builds submit` needs to add this substitution as well.

tmatsuo: You're right. I finally see the meaningful error message when debug is on.

tmatsuo:
Now the test should pass (in theory)

tmatsuo:
@chingor13 Can we merge?

chingor13:
Yep looks good

PR #351, opened 2017-08-18
Allow individual packages for stackdriver integration

tmatsuo:


tmatsuo:
cc @bshaffer

chingor13:
Do we know what happens when both `google/cloud` and `google/cloud-logging` are installed? Do we have it set up for the individual package to provide logging capabilities instead of `google/cloud`'s?

tmatsuo:
@chingor13 
`google/cloud` has [`replace` field](https://github.com/GoogleCloudPlatform/google-cloud-php/blob/master/composer.json#L65).

I don't think you can install both `google/cloud` and `google/cloud-logging`.

Adding @jdpedrie to confirm

jdpedrie:
@tmatsuo @chingor13 that is correct. You can't have both installed at the same time. However, if you require both `google/cloud` and `google/cloud-logging`, the latter should be [replaced by a compatible version of `google/cloud`](https://github.com/GoogleCloudPlatform/google-cloud-php/blob/master/composer.json#L70). ([more information](https://getcomposer.org/doc/04-schema.md#replace))

tmatsuo:
Yeah, my plan is to add a new integration test with single packages.

tmatsuo:
@chingor13 Added the test app.

We can run the integration tests before the release to make sure it all works :)

PR #359, opened 2017-09-11
Install opencensus extension

tmatsuo:


chingor13:
We can add a basic extension loaded test as well.

testapps/build_pipeline/tests/tests/EndToEndTest.php
```diff
@@ -198,6 +198,28 @@ public function testGoodbye()
         });
     }
 
+    /**
+     * @dataProvider extensions
+     */
+    public function testExtensions($extensionName)
+    {
+        $this->runEventuallyConsistentTest(function () {
```
chingor13: ah, this needs a `use ($extensionName)` to bind the variable into the closure.

PR #363, opened 2017-09-12
Remove /app from the config-scan-dir

### LINKED ISSUE #361: Remove /app from the config scan dir
It causes #360

tmatsuo:
fixes #361


tmatsuo:
Note: Unfortunately the test result for this particular PR in circle is not testing this change.
We will be able to test it after creating the debian packages with this change.

Issue #366, opened 2017-09-14
Allow overriding composer install flags


Matan:
Hi!

I'd like to be able to override the `composer install` flags passed here: https://github.com/GoogleCloudPlatform/php-docker/blob/master/php-base/build-scripts/composer.sh#L88-L93

I'm creating 3 environments (projects) on GCP namely `dev`, `stage` and `prod`. I need to be able to load the development version of application based on the docker build. Unfortunately there is no way for me to keep the development composer packages as `--no-dev` is hard coded into the build process.

Keep these static:

```
--optimize-autoloader
--no-interaction
--no-ansi
--no-progress
```

Then replace these:

```
--no-dev
--prefer-dist
```

with a build arg defaulted to these sensible values.

Thanks!

tmatsuo:
@Matan Thanks for the feedback.

Which base image do you use, gcr.io/google-appengine/php, or versioned one like gcr.io/google-appengine/php71?

At a first glance, yeah it does make sense to introduce a build arg {`COMPOSER_FLAGS`} defaults to `--no-script --no-dev --prefer-dist` or `--no-dev --prefer-dist` depending on the build pipeline.

Matan:
Hi! I'm using the `gcr.io/google_appengine/php` base image. During build it detects 7.1 requirement correctly.

```
...
---> Running in 9db501558c3e
Detecting PHP version...
Using PHP version: 7.1
Install PHP extensions...
...
```

tmatsuo:
@Matan Thanks you're inheriting from `gcr.io/google_appengine/php`, right? Is it possible to show your Dockerfile?

Matan:
Sure, here it is:

```
# The Google App Engine php runtime is Debian Jessie with PHP installed
# and various os-level packages to allow installation of popular PHP
# libraries. The source is on github at:
#   https://github.com/GoogleCloudPlatform/php-docker
FROM gcr.io/google_appengine/php

# Add our NGINX and php.ini config
ENV DOCUMENT_ROOT=${APP_DIR}/web \
    NGINX_CONF_INCLUDE=conf/nginx.conf \
    PHP_INI_OVERRIDE=conf/php.ini

# Manually create parameters.yml
RUN rm -f ${APP_DIR}/app/config/parameters.yml \
 && mv ${APP_DIR}/app/config/parameters.yml.dist ${APP_DIR}/app/config/parameters.yml
```

[...]

PR #368, opened 2017-09-14
Customizable composer flags

tmatsuo:
See #366 for discussion

### DIFF
--- a/php-base/build-scripts/composer.sh
+++ b/php-base/build-scripts/composer.sh
@@ -78,7 +78,7 @@ EOF
     echo "Running composer..."
     # Run Composer.
     if [ -z "${COMPOSER_FLAGS}" ]; then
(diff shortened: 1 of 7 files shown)


php-base/build-scripts/composer.sh
```diff
@@ -77,18 +77,15 @@ EOF
...
-          install ${NOSCRIPT} \
-          --no-dev \
-          --prefer-dist \
+          install \
           --optimize-autoloader \
           --no-interaction \
           --no-ansi \
-          --no-progress"
+          --no-progress \
+          ${COMPOSER_FLAGS}"
```
chingor13: Will this be broken if COMPOSER_FLAGS is empty?

chingor13: nevermind, I guess the default is being set above.

tmatsuo:
Oops replaced `--no-script` with `--no-scripts` which should fix the test.

PR #371, opened 2017-09-20
Use larger client_body_buffer_size for nginx

### LINKED ISSUE #228: Set client_body_buffer_size to appropriate value
The default is 8k which is small.
How about to raise it to say, 128k?

tmatsuo:
will fix #228


tmatsuo:
We talked offline. We decided to defer this config to users.

PR #372, opened 2017-09-20
Checks the nginx config file syntax after moving user supplied config

### LINKED ISSUE #268: Deploy doesn't report bad nginx config.
Unfortunately I seem to have had a lot of experience writing nginx config that is broken, but when deploying to Flex it isn't picking up the error with the config and stopping the deployment, although the HTTP health-checks are failing so it doesn't switch to the new version and the error is logged in the stderr log, the command just hangs for a very long time and you have to go check the logs manually.

Speaking with support (Case 12563159), they said for the [example nginx container](https://github.com/GoogleCloudPlatform/appengine-custom-runtimes-samples/tree/master/nginx) the deployment would halt quite quickly with the nginx error message spat out to the console of the deployer, but it looks like because nginx is running via supervisord that functionality doesn't extend to this runtime.

tmatsuo:
will fix #268

### DIFF
--- a/php-base/build-scripts/entrypoint.sh
+++ b/php-base/build-scripts/entrypoint.sh
@@ -22,6 +22,9 @@ set -xe
 
 /bin/bash /build-scripts/move-config-files.sh
 
+# Checks the nginx config. We need to do this after moving the files.
+/usr/sbin/nginx -t -c /etc/nginx/nginx.conf
+
 # Configure memcached based session.
 if [ -n "${MEMCACHE_PORT_11211_TCP_ADDR}" ] && [ -n "${MEMCACHE_PORT_11211_TCP_PORT}" ]; then
     cat <<EOF > ${PHP_DIR}/lib/conf.d/memcached-session.ini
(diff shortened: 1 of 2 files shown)


php-base/build-scripts/entrypoint.sh
```diff
@@ -22,6 +22,9 @@ set -xe
 
 /bin/bash /build-scripts/move-config-files.sh
 
+# Checks the nginx config. We need to do this after moving the files.
+/usr/sbin/nginx -t -c /etc/nginx/nginx.conf
```
chingor13: Do we actually need to move the config files again in this script (and thus check the nginx.conf again)?

It might have been left over from the xrt deployments when should no longer be used (unless we're still supporting them).

tmatsuo: Users are still using the `gcr.io/google-appengine/php` image, because it is the only way to customize the php runtime right now.

You can not do the same check in your customized `Dockerfile` because the config files are not there yet.

PR #373, opened 2017-09-21
Added how to enable OPcache

### LINKED ISSUE #369: Opcache is not enabled by default.
Opcache extension is not loaded even though the document says it is enabled by default.
i had to add `zend_extension=opcache.so` to `php.ini`  to enable opcache.

tmatsuo:
fixes #369


tmatsuo:
@dpebot merge when green

PR #375, opened 2017-09-25
Use `cloudbuild` network for accessing metadata server

tmatsuo:
Instead of setting an envvar `GCLOUD_PROJECT` which may be deprecated in the future, it is better to use the `cloudbuild` network for accessing metadata server.

CC: @skelterjohn
Is there any plans for having a public documentation about this?


tmatsuo:
cc @bshaffer

PR #378, opened 2017-10-03
Allow building deprecated mongo from pecl

jaitaiwan:
The mongo pecl extension is needed for phalcon as the object document
mapper provided by phalcon works exclusively with the deprecated library. This change includes the extension and changes the readme to show that it is now allowed.


jaitaiwan:
Running CLA Check Again

deb-package-builder/extensions/mongo/debian/copyright
```diff
@@ -0,0 +1,72 @@
+Format: http://www.debian.org/doc/packaging-manuals/copyright-format/1.0/
+Upstream-Name: mongo
+Upstream-Contact: Daniel Holmes <dan@djcentric.com>
+Source: https://pecl.php.net/package/mongo
+
+Files: *
```
tmatsuo: According to https://pecl.php.net/package/mongo, it seems like the Apache License

jaitaiwan: Source code shows Apache 2.0. Is there something that needs changing here?

jaitaiwan:
I should mention that it didn't look like this extension was built when I tried this in my own local env. What's the best way to test that this works without waiting to rebuild all the packages?

deb-package-builder/extensions/mongo/debian/copyright
chingor13: Below this line specifies the copyright holder and the license. It looks like the copyright holder should be `Copyright 2009-2014 MongoDB, Inc.` and the license should be `Apache-2.0`

chingor13:
Generally to test the extension building scripts, I build the docker image locally and run it.

```
cd deb-package-builder
docker build -t deb-package-builder .
docker run --rm -v $(pwd):/workspace 5.6.31-1 mongo
```

It will take awhile to compile PHP, but the compiled .deb file will be cached locally in the `pkg` folder for future runs.

jaitaiwan:
Things are a bit delayed on my side of the fence as this is not a high priority. Will get around to it soon though I expect

jaitaiwan:
@chingor13 or another contributor. Can someone point me to how to test the php-docker image after building the package?

chingor13:
@jaitaiwan Our automated testing framework tests packages once they are pushed to our staging apt package server, so that won't work for PRs.

To test locally, what I do is follow the steps in [this comment](https://github.com/GoogleCloudPlatform/php-docker/pull/378#issuecomment-333891816) to build a `.deb` file. Then I create a sample application that builds in docker with something like:

```
FROM gcr.io/google-appengine/php
ADD * /app
RUN dpkg -i /app/[filename].deb
```
Then I run that built image (with tests) or inspect from the command line.

jaitaiwan:
Not sure what is going on with the google bot... The CLA is associated with the email that I used for this commit...

jaitaiwan:
@tmatsuo @chingor13 Thanks for your assistance. I've tested this locally and it seems to work perfectly as far as I can tell.

deb-package-builder/extensions/mongo/debian/copyright
```diff
@@ -0,0 +1,66 @@
...
+Copyright: Copyright 2009-2014 MongoDB, Inc.
+License: APACHE-2.0
```
chingor13: nit: change to `Apache-2.0` to match other examples.

deb-package-builder/extensions/mongo/debian/copyright
```diff
@@ -0,0 +1,66 @@
...
+  Apache License
```
chingor13: For well-know license types, we don't need to include the text of the license. See: the example from the debian files below the main source in this same file.

deb-package-builder/extensions/mongo/build.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+set -ex
+source ${DEB_BUILDER_DIR}/functions.sh
+
+echo "Building mongo for gcp-php${SHORT_VERSION}"
+
+PNAME="gcp-php${SHORT_VERSION}-mongo"
+
+# Download the source
+if [ ${SHORT_VERSION} == '70' ] || [ ${SHORT_VERSION} == '71' ]; then
```
chingor13: Maybe reverse this to only build for PHP 5.6? We also build extensions for 7.2 as well, so this will likely crash for 7.2 builds.

jaitaiwan:
Thanks for the feedback. Will make these changes and have it ready in a few minutes

tmatsuo:
@jaitaiwan FYI, we'll make another release later this week (probably on Thu).

jaitaiwan:
@chingor13 requested changes made.  @tmatsuo If no more reviews this is ready :)

php-base/build-scripts/move-config-files.sh
```diff
@@ -47,15 +47,15 @@ fi
 
 # Move user-provided nginx config files.
 if [ -f "${NGINX_CONF_INCLUDE}" ]; then
-    mv "${NGINX_CONF_INCLUDE}" "${NGINX_USER_CONF_DIR}/nginx-app.conf"
```
tmatsuo: Strictly saying, I would separate this change from this PR because it's unrelated.

chingor13:
@jaitaiwan The [new runtime](https://github.com/GoogleCloudPlatform/php-docker/releases/tag/2018-02-23-15-30) that includes the mongo extension for 5.6 should now be available.

PR #379, opened 2017-10-17
Use expectException instead

tmatsuo:
Somehow `setExpectedExceptionRegExp` changed the bahavior with the phpunit 5.7.23.
I think it's safer to use `expectException`.


tmatsuo:
Our nightly test started to fail, and I got repro with phpunit 5.7.23. This PR fixed the test failure locally and it should pass on circleCI too.

tmatsuo:
@dpebot merge when green

PR #385, opened 2017-11-12
Use PHPUnit\Framework\TestCase instead of PHPUnit_Framework_TestCase

gabriel-caruso:
I use the `PHPUnit\Framework\TestCase` notation instead of `PHPUnit_Framework_TestCase` while extending our TestCase. This will help us migrate to PHPUnit 6, that [no longer support snake case class names](https://github.com/sebastianbergmann/phpunit/blob/master/ChangeLog-6.0.md#changed-1).

I just need to bump PHPUnit version to [`4.8.35`](https://github.com/sebastianbergmann/phpunit/blob/master/ChangeLog-4.8.md#4835---2017-02-06), that support this namespace.

### DIFF
--- a/builder/gen-dockerfile/tests/GenFilesCommandTest.php
+++ b/builder/gen-dockerfile/tests/GenFilesCommandTest.php
@@ -17,8 +17,9 @@
 namespace Google\Cloud\Runtimes\Builder;
 
 use Symfony\Component\Console\Tester\CommandTester;
+use PHPUnit\Framework\TestCase;
 
-class GenFilesCommandTest extends \PHPUnit_Framework_TestCase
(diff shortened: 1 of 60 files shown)


tmatsuo:
@gabriel-caruso Thanks for your contribution!
LGTM

@chingor13 Can you look at it too just in case?

tmatsuo:
Maybe we should create another PR, preserving the commit by @gabriel-caruso so that we can run the test on CircleCI?

chingor13:
Pushed this branch to origin as `phpunit-fork` so CircleCI is running the tests.

chingor13:
Thanks!

PR #389, opened 2017-11-20
Allow overriding the URL for downloading pre ga source packages

tmatsuo:


tmatsuo:
Locally it successfully built the 7.2.0RC6-1 deb package

PR #391, opened 2017-11-30
Making the default to PHP 7.2

tmatsuo:


tmatsuo:
The test failed with:

```
PHP Warning: copy(): SSL: Connection reset by peer in Command line code on line 1
echo 'ERROR: Invalid composer installer signature'
ERROR: Invalid composer installer signature
```
Likely a temporary errror, but better to making that part more robust.

PR #395, opened 2017-12-06
Integration test fixes for PHP 7.2

tmatsuo:


tmatsuo:
It's better to run the integration test on CircleCI too, #396 filed

tmatsuo:
For now the integration tests are passing locally

PR #397, opened 2017-12-11
Allow versions like 3.5.0.1 for extension debian packages

tmatsuo:


deb-package-builder/functions.sh
```diff
@@ -47,7 +47,7 @@ download_from_pecl()
         with_retry "pecl download ${PECL_PACKAGE_NAME}" 6
         # determine the downloaded version
         EXT_VERSION=$(ls ${PACKAGE_SHORT_NAME}-*.tgz | \
-                sed "s/${PACKAGE_SHORT_NAME}-\([0-9][0-9]*\.[0-9][0-9]*\.[0-9][0-9]*\)\.tgz/\1/")
+                sed -E "s/${PACKAGE_SHORT_NAME}-([0-9]+.[0-9]+(.[0-9]+)+).tgz/\1/")
```
chingor13: Do we need to keep the escaping for the "." characters?

tmatsuo: Good catch, done

tmatsuo:
Just in case I'm running the package builder locally. I'll merge once it builds correct packages

tmatsuo:
Now it's failing on opencensus-0.1.0RC1.tgz

chingor13:
Doh, I uploaded a devel version of opencensus to ensure the build works. 0.1.0 will be tagged alpha and will be downloadable via `pecl download opencensus-alpha`.

In any case, we should probably make the version string even more lenient.

tmatsuo:
Yeah I'm thinking to allow any postfix string

tmatsuo:
It builds correct packages including
- gcp-php72-opencensus_0.1.0RC1-7.2.0-2_amd64.deb
- gcp-php72-protobuf_3.5.0.1-7.2.0-2_amd64.deb

PR #398, opened 2017-12-12
Use opencensus alpha

tmatsuo:
also use stackdriver_debugger devel channel


tmatsuo:
@chingor13 OK to merge?

PR #401, opened 2017-12-13
Remove libip2location

tmatsuo:


tmatsuo:
I'll merge this for now. Maybe we can check the changed file and only run relevant tests in the future.

PR #406, opened 2018-01-05
Adding Sara's key to the php71 directory temporarily.

tmatsuo:


tmatsuo:
Ok, changed the commit message and it should be fine

Issue #409, opened 2018-01-09
Sodium extension not included in PHP 7.2 image


chingor13:

chingor13:
Fixed in https://github.com/GoogleCloudPlatform/php-docker/releases/tag/2018-01-09-13-25

PR #410, opened 2018-01-09
Compile PHP 7.2 with-sodium

chingor13:
See #409


chingor13:
Testing the compilation locally. We can't test the new version with these integration tests until we build the new debian package for the cloud apt repository.

chingor13:
This compiles locally and includes the sodium extension.

Issue #426, opened 2018-03-27
Stackdriver Integration version detection too strict


chingor13:
with an `app.yaml` containing `enable_stackdriver_integration: true`, we do not allow a wildcard version for `google/cloud`, `google/cloud-logging`, or `google/cloud-error-reporting`.

Additionally, if a `composer.lock` file is present, we should respect the locked version when doing our version comparison.

tmatsuo:
Yeah. Right now, it fails all the constraints without the lower bound.

PR #428, opened 2018-03-27
Change the way we check the required version of the dependency

tmatsuo:
There are still some cases for false negative where the direct
constrant is ok and other indirect dependency has a problem.

Although the issue is not completely fixed, this logic change allows
more legitimate cases which has been failing with the previous code.

It will alleviate #426


tmatsuo:
For a long time fix, the best way is to invoke `composer install` first, then look at the `composer.lock` file to know the exact version the app is using. This will need non-trivial changes to the builder pipeline. It might make sense to put this band-aid on while we're working on the long-term fix. WDYT?

PR #437, opened 2018-05-02
Build ubuntu images

tmatsuo:
The PR originally consists of 2 commits. The first commit is just copying `php-base` to `ubuntu-php-base`, `cloudbuild.yaml` to `cloudbuild-ubuntu.yaml`. See the 2nd commits and the after for the actual changes.


tmatsuo:
Oh, I think we need to use the ubuntu-based images in the e2e test too. I'll work on that.

ubuntu-php-base/Dockerfile
```diff
@@ -65,15 +68,15 @@ COPY build-scripts /build-scripts
 # Files for stackdriver setup
 COPY stackdriver-files /stackdriver-files
 
-RUN chown www-data /build-scripts
+RUN chown -R www-data /build-scripts /stackdriver-files
 
-ARG RUNTIME_DISTRIBUTION="gcp-php-runtime-jessie"
+ARG RUNTIME_DISTRIBUTION="gcp-php-runtime-xenial-20180425-1"
```
chingor13: Should the default be `gcp-php-runtime-xenial-unstable`?

cloudbuild-ubuntu.yaml
```diff
@@ -282,11 +263,10 @@ steps:
     id: php72-extensions-test
 
 images:
-  - gcr.io/${_GOOGLE_PROJECT_ID}/php-base:$_TAG
-  - gcr.io/${_GOOGLE_PROJECT_ID}/php72:$_TAG
-  - gcr.io/${_GOOGLE_PROJECT_ID}/php71:$_TAG
-  - gcr.io/${_GOOGLE_PROJECT_ID}/php70:$_TAG
-  - gcr.io/${_GOOGLE_PROJECT_ID}/php56:$_TAG
-  - gcr.io/${_GOOGLE_PROJECT_ID}/php:$_TAG
-  - gcr.io/${_GOOGLE_PROJECT_ID}/php/gen-dockerfile:$_TAG
```
chingor13: We should probably still have a gen-dockerfile ubuntu image?

cloudbuild-ubuntu.yaml
```diff
@@ -7,102 +7,83 @@ steps:
...
     dir: php-onbuild
     waitFor: ['php71']
     id: php-onbuild
   - name: gcr.io/gcp-runtimes/structure_test
-    args: ['-i', 'gcr.io/${_GOOGLE_PROJECT_ID}/php:$_TAG', '--config', '/workspace/php-onbuild/php.yaml', '-v']
+    args: ['-i', 'gcr.io/${_GOOGLE_PROJECT_ID}/ubuntu-php:$_TAG', '--config', '/workspace/php-onbuild/php.yaml', '-v']
     waitFor: ['php-onbuild']
     id: php-onbuild-structure
 
-  # gen-dockerfile
```
chingor13: Is there a reason we're not creating the gen-dockerfile image?

tmatsuo:
`gcr.io/google-appengine/ubuntu-php72`

Few thoughts on the image name.

I think we may want to use other project name like `gcp-runtimes`
For example, how about the following pattern?
`gcr.io/gcp-runtimes/jessie-php72`
`gcr.io/gcp-runtimes/xenial-php72`
etc?

If we keep the `google-appengine` project name, I prefer keeping `php` as it is becaues, in terms of app engine, the image name doesn't matter. For other use cases, I'd prefer keeping the same behavior with the same image names.

WDYT?

tmatsuo:
I see. Let's do that. I think we can change the `google-appengine/php` to ubuntu based images when we switch the default :)

PR #438, opened 2018-05-07
Makes the ubuntu image smaller

tmatsuo:


ubuntu-php-base/build-scripts/install_php72.sh
```diff
@@ -18,27 +18,16 @@
...
-        gcp-php72-oauth \
-        gcp-php72-opencensus \
-        gcp-php72-pq \
-        gcp-php72-protobuf \
-        gcp-php72-rdkafka \
-        gcp-php72-redis \
-        gcp-php72-stackdriver-debugger \
-        libsodium18 \
-        --no-install-recommends
+        libsodium18
```
chingor13: I don't think we need this anymore. In the first release, we forgot to add the dependency to `gcp-php72` but it should be present now.

tmatsuo: Removed

PR #439, opened 2018-05-08
Added base_os option in the runtime_config

tmatsuo:


testapps/ubuntu_php72/logging-client.php
```diff
@@ -0,0 +1,30 @@
+<?php
+/*
+ * Copyright 2017 Google Inc.
```
chingor13: nit: should these copyright years be 2018?

testapps/ubuntu_php72/composer.lock
```diff
@@ -0,0 +1,1491 @@
...
+        }
+    ],
+    "packages-dev": [],
+    "aliases": [],
+    "minimum-stability": "stable",
+    "stability-flags": [],
+    "prefer-stable": false,
+    "prefer-lowest": false,
+    "platform": {
+        "php": "7.1.*",
```
chingor13: This PHP version doesn't match the `composer.json` version.

PR #441, opened 2018-05-11
Record deployment latency for both ubuntu16 and debian8 images

tmatsuo:


tmatsuo:
It's locally working with my test project

perf-dashboard/deployment-latency/tests/CollectDeploymentLatencyTest.php
```diff
@@ -58,49 +57,66 @@ public function testDeploymentLatency()
...
+                try {
+                    $reportName = sprintf('%s-%s', $type, $phpVersion);
+                    $failureCount = 0;
+                    $command = 'gcloud -q ' . $gcloudTrack . ' app deploy'
+                        . ' --version '
+                        . str_replace('.', '', $reportName)
+                        . ' --no-stop-previous-version --no-promote';
+                    $configCmd = 'gcloud config set app/use_runtime_builders true';
+                    self::execWithError($configCmd, 'runtime-builders-config');
+                    $f = fopen('app.yaml', 'a+');
```
chingor13: won't there be two `base_os` lines in the second run through if we're creating the app outside of the `$types` foreach?

tmatsuo: We're creating the app inside of the foreach now

PR #448, opened 2018-06-06
Push the gen-dockerfile image after building in cloudbuild

chingor13:
The debian gen-dockerfile image was being used for both Debian and Ubuntu pipelines. We need to push the ubuntu on in the ubuntu-cloudbuild.yaml


chingor13:
Tests were passing previously because the image was tagged on the local docker/cloudbuild instance, but not available outside that build.

Issue #453, opened 2018-07-23
Phalcon for PHP 7.1 & 7.2


TriPSs:
Phalcon supports php 7.2 from release 3.3.

https://github.com/phalcon/cphalcon/blob/master/CHANGELOG-3.3.md#330-2017-12-23

Is it possible to remove the restrictions that Phalcon can only be installed on php 7.0

jaitaiwan:
I'm just leaving a comment here because I can handle this sometime this week.

jaitaiwan:
@TriPSs I've added a pull request in. Hopefully won't be long till feedback/acceptance and a new version is put live.

jaitaiwan:
@sharifelgamal This issue can be closed.
Happy to become a maintainer if there's help needed.

TriPSs:
Thanks @jaitaiwan!

PR #454, opened 2018-07-23
Removed Phalcon from unavailable extensions

TriPSs:
Phalcon supports PHP 7.1 and 7.2 as of v3.3.0

Issue: #453


TriPSs:
I signed it!

TriPSs:
Noticed that more work is required to let it work

PR #462, opened 2018-10-29
Pin libsodium to 2.0.12

tmatsuo:


tmatsuo:
@chingor13 Seems like libvips changed its download URL and I needed to correct it

tmatsuo:
Ok, I successfully built the packages locally. Merging this

PR #471, opened 2019-03-13
Enable opcache file to improve CLI scripts speed

ibrunotome:
As all the GAE deployments are atomic, we can compile with the flag `--enable-opcache-file`, and the user choose if he wants or not to enable it on the php.ini file.


ibrunotome:
I signed it!

PR #482, opened 2019-07-09
gsutil is broken, using gcloud-sdk image while waiting fix

donmccasland:


donmccasland:
Here's a description of the breakage: https://github.com/GoogleCloudPlatform/gsutil/issues/806
and the proposed workaround for cloudbuild:
https://github.com/GoogleCloudPlatform/cloud-builders/issues/508#issuecomment-507646005

PR #491, opened 2019-10-24
Fix ubuntu package builder

donmccasland:


package-builder/Dockerfile
```diff
@@ -100,7 +100,14 @@ RUN apt-get update -y && \
         fakeroot \
         libtool \
         automake \
-        autotools-dev
+        autotools-dev \
+	wget \
```
tejal29: small nit: space indentation is off.

PR #492, opened 2019-10-28
Adding librdkafka to php-base

donmccasland:


donmccasland:
Adding a comment.  The 4.0.0 version of rdkakfa php extension requires lib-rdkafka-dev version > 0.11

PR #519, opened 2020-07-22
Fixing tests for 7.3

donmccasland:
Removing composer.lock files that are automatically generated during build, but foul up tests if they are part of the repo.
Also SERVICE_ACCOUNT_JSON is no longer required, although it can be set and used.


johnhout:
So when will this be tagged?

PR #544, opened 2021-10-08
Removing 7.4 and 8.0 images from builder until we can delivery them

donmccasland:


jinglundong (approved):

Stoyvo:
Can the public help resolve this? PHP 7.4 is critical to have support for, PHP 8.0 is already 1 release behind.

